PR #2, opened 2023-01-30
update branch references and remove connectors build

cpdeethree:


cpdeethree:
The platform ci test is flaky due to an unrelated reason that still exists in the codebase. Tim put a fix up here that we can cherry pick in: https://github.com/airbytehq/airbyte/pull/22087

PR #11, opened 2023-02-17
Remove AirbyteGitHubStore in favor of the remote oss catalog

bnchrch:
## What
This is part of a series of changes (#22653) to refactor platform to no longer reference the YAML files directly.
This is a continuation of this PR https://github.com/airbytehq/airbyte/pull/23055#issuecomment-1433739415
## How
This PR replaces the implementation of AirbyteGitHubStore with one that references the remote oss_catalog

## Notes for reviewers
1. I explored converting this to a explicitly singleton class (similar to RemoteDefinitionProvider) but unfortunately cloud makes explicit calls to `SourceDefinitionsHandler` [outside of micronaut](https://github.com/airbytehq/airbyte-platform-internal/blob/master/airbyte-server-wrapped/src/main/java/io/airbyte/server/wrapped/ConfigurationApiWrapped.java#L287) cc: @colesnodgrass 
2. I would like to have this use the RemoteDefinitionProvider singleton, however we need this to always go to OSS definitions not cloud definitions


airbyte-commons-server/src/test/java/io/airbyte/commons/server/services/AirbyteGithubStoreTest.java
```diff
@@ -1,161 +0,0 @@
-/*
- * Copyright (c) 2023 Airbyte, Inc., all rights reserved.
- */
-
-package io.airbyte.commons.server.services;
```
bnchrch, line 5: Note: I removed these tests as the RemoteDefinitionProvider tests cover this class.

Let me know if I should still test this specific class further.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendCheckUpdatesHandler.java
```diff
@@ -32,11 +32,11 @@ public class WebBackendCheckUpdatesHandler {
   // todo (cgardens) - this handler should NOT have access to the db. only access via handler.
   @Deprecated
   final ConfigRepository configRepositoryDoNotUse;
-  final AirbyteGithubStore githubStore;
+  final AirbyteRemoteOssCatalog remoteOssCatalog;
 
-  public WebBackendCheckUpdatesHandler(final ConfigRepository configRepositoryDoNotUse, final AirbyteGithubStore githubStore) {
+  public WebBackendCheckUpdatesHandler(final ConfigRepository configRepositoryDoNotUse, final AirbyteRemoteOssCatalog remoteOssCatalog) {
```
pedroslopez, line 37: Just for my understanding, where/when is this WebBackendCheckUpdatesHandler used?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteRemoteOssCatalog.java
```diff
@@ -0,0 +1,28 @@
...
+  private static final String REMOTE_OSS_CATALOG_URL = CatalogDefinitionsConfig.getRemoteOssCatalogUrl();
+  private static final long TIMEOUT = 30000;
+  private static final Logger LOGGER = LoggerFactory.getLogger(CombinedConnectorCatalogDownloader.class);
+
+  public AirbyteRemoteOssCatalog() {
+    super(REMOTE_OSS_CATALOG_URL, TIMEOUT);
+    LOGGER.info("Initializing OSS catalog from {} with timeout {}", REMOTE_OSS_CATALOG_URL, TIMEOUT);
+  }
+
+}
```
pedroslopez, line 28: > I would like to have this use the RemoteDefinitionProvider singleton, however we need this to always go to OSS definitions not cloud definitions

Could you elaborate on why we need this to always go to OSS definitions? The RemoteDefinitionsProvider in use should point to the right one depending on the environment (cloud to cloud catalog, oss to oss catalog). If we ever decide to do somehting on cloud to show the latest versions, I think it makes sense for it to use the Cloud catalog as source of truth for that rather than the OSS catalog. We were hardcoding it to the YAMLs available in OSS before because these methods were only being called in OSS, but it doesn't feel like a requirement to me.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandler.java
```diff
@@ -152,11 +152,7 @@ public SourceDefinitionReadList listLatestSourceDefinitions() {
   }
 
   private List<StandardSourceDefinition> getLatestSources() {
-    try {
-      return githubStore.getLatestSources();
-    } catch (final InterruptedException e) {
-      throw new InternalServerKnownException("Request to retrieve latest destination definitions failed", e);
-    }
+    return remoteOssCatalog.getSourceDefinitions();
```
pedroslopez, line 155: Do we need to set up micronaut caching on the server app for this to get the latest values? `getRemoteDefinitionCatalog` is marked as cacheable, which I believe by default if there's no configuration means that we'll only retrieve the values once and cache the results essentially until server restart. 

airbyte-cron has an example for how the config is modified there to cache the results for 15 seconds:

https://github.com/airbytehq/airbyte-platform/blob/016e80f82a7027af7840ba20689f57fabb853c8c/airbyte-cron/src/main/resources/application.yml#L4-L6

pedroslopez (changes requested):
Nice! Only blocking comment is around caching, but I also think we should be able to use RemoteDefinitionsProvider directly without going through the AirbyteRemoteOssCatalog

airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteRemoteOssCatalog.java
bnchrch: @pedroslopez So I was going off of a comment left by Evan that I had misintterpretted.

https://github.com/airbytehq/airbyte/pull/23055#issuecomment-1433739415

I had originally understood it as we always want to point to OSS even in cloud (which is what we have been doing)

But after rereading it this morning its seems clear that what we are currently doing is not correct and we do want to reference the cloud catalog for this feature. Which means youre right we can remove this and just reference the DefinitionProvider

@evantahler Triple checking, is that correct?

evantahler: I think I'm confused.  Shouldn't the OSS platform use REMOTE_OSS_CATALOG_URL and the cloud platform use REMOTE_CLOUD_CATALOG_URL?

bnchrch: Ok so this IS confusing, not just you.

For context sake, in the video below I try to explain
1. How this was working before the split
2. The assumptions/context as to why
3. How its been designed to work during the split

This has led to two questions for the product (@evantahler )
1. Should we continue to use `LocalDefinitionsProvider` in oss to support airgapped? (Like we have been)
2. `AirbyteGithubStore` always referenced the OSS version of the catalog, regardless of the context it was running in (oss/cloud), should we continue that behaviour? Or should we just use the same catalog used to seed the database?

https://www.loom.com/share/cbe547ead2b44f0086a8066c34280b30

pedroslopez: (Sweet loom!) My take:

1. Yes, we should continue to use LocalDefinitionsProvider in OSS. We can decide later on to bring in auto-updates and remote definitions to load connectors in OSS, but I think that's orthogonal to this work and a different conversation.
2. We should just use the same catalog used to seed the database. This was "a bug", but in practice this hasn't been an issue in the past because it's important to realize *users can't upgrade connectors on their own for cloud*. This also means we don't show the latest version on that settings page in Cloud, unlike we do in OSS.

bnchrch: Another loom reply!
https://www.loom.com/share/e05991c0251d46569cd93e9020f41b05

Basically I agree 100% with Pedro but I think to change to the DI version of `DefinitionProvider` will require a larger refactor as we're currently between a migration to micronaut?

I believe this is what @colesnodgrass and I talked about.

pedroslopez: Thanks again for the loom!

Argh yeah, I think the in-progress micronaut migration makes this a bit harder. I'm ok with pushing this piece off until the cloud server has been migrated to micronaut, which seems like it's very close anyway. 

Regardless, I think something else you said in that loom makes me think we _shouldn't_ actually use the same global DefinitionsProvider singleton, and that's the fact that we would be using LocalDefinitionsProvider to get latest versions in OSS since we're using LocalDefinitionsProvider to load things in... _but_ we want to use a RemoteDefinitionsProvider instead to get the lateset versions. This means that we would have a LocalDefinitionsProvider (for initial loading) _and_ a RemoteDefinitionsProvider (for latest versions). We may be able to still use RemoteDefinitionsProvider directly for this, but set it up as a different "named" bean, but I don't think you'll be able to do this until the server has been moved to micronaut, anyway.

All this to say, I think we might have to keep a separate class for this for now.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandler.java
bnchrch: @pedroslopez I think by definition its cached as `RemoteDefinitionsCatalog:getRemoteDefinitionCatalog` is tagged with `@cachable`

airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteRemoteOssCatalog.java
bnchrch: Ok! Well silver lining for me here is that I didn't misunderstand the system!

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendCheckUpdatesHandler.java
bnchrch: Looks like its used in an api controller https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-server/src/main/java/io/airbyte/server/apis/WebBackendApiController.java#L46

airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteRemoteOssCatalog.java
bnchrch: Added a @deprectaed decorator to the class to mark this for removal on micronaut migration

airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteRemoteOssCatalog.java
```diff
@@ -0,0 +1,30 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Convenience class for retrieving the remote OSS Catalog.
+ */
+// This should be deleted, in favour of the DefinitionsProvider Singleton when cloud is migrated to
+// micronaut
+@Deprecated(forRemoval = true)
+@SuppressWarnings("PMD.AvoidCatchingThrowable")
```
pedroslopez, line 20: Just curious: What is this suppressing?

airbyte-server/src/main/java/io/airbyte/server/config/ApplicationBeanFactory.java
```diff
@@ -98,8 +98,8 @@ public JsonSchemaValidator jsonSchemaValidator() {
   }
 
   @Singleton
-  public AirbyteGithubStore airbyteGithubStore() {
-    return AirbyteGithubStore.production();
+  public AirbyteRemoteOssCatalog remoteOssCatalog() {
+    return new AirbyteRemoteOssCatalog();
   }
```
pedroslopez, line 103: I believe if there's no arguments to pass this can be removed as long as the AirbyteRemoteOssCatalog is marked as `@Singleton` itself

pedroslopez (commented):
Code LGTM, but could we make sure this works functionally by maybe pointing it to a local server and making some changes after the server's already run it?

I still think the default caching that's in place will mean we only get the values once on first call and then never get updates after the fact. I believe we'll need to either specify a cache expriation in airbyte-server's application.yml as is being done for airbyte-cron, or maybe we can set these defaults on the RemoteDefinitionProvider class itself.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandler.java
pedroslopez: Yes, but my concern is that I believe this default means that the cache never expires, so in practice we would only fetch what the latest version is _once_, and then always reuse the same values and never make another request to fetch latest values again.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteRemoteOssCatalog.java
bnchrch: I am not sure..

This was from the old class.

ill remove, see what happens.

pedroslopez (approved):
Ben and I tested this out and caching seems to be working fine!

:shipit:

PR #12, opened 2023-02-17
Remove the report-connectors-dependency.yml workflow

bnchrch:
## Overview
Remove the report-connectors-dependency.yml workflow as it is specifc to integrations

## Notes for reviewer
Sanity check, the version in the oss repo should still work as expected.


erohmensing (changes requested):
I think we should also remove the scripts called by this file and any files they use:
* `python ./tools/bin/ci_check_dependency.py ./changed_files.txt`
* `connector_dependency_template.md`

after of course sanity checking that they're not used anywhere else

cpdeethree:
> I think we should also remove the scripts called by this file and any files they use:
> 
> * `python ./tools/bin/ci_check_dependency.py ./changed_files.txt`
> * `connector_dependency_template.md`
> 
> after of course sanity checking that they're not used anywhere else

I believe I remove the template in https://github.com/airbytehq/airbyte-platform/pull/9

But need to double check. I don't touch anything in /tools/bin in that PR

cpdeethree:
I'll have a new PR up today that cleans out /tools/bin as well

cpdeethree:
https://github.com/airbytehq/airbyte-platform/pull/13

PR #14, opened 2023-02-17
delete local connector catalogs

cpdeethree:
## What
Removes the local connector catalogs

## How
Removes the files

This build probably won't succeed until the remainder of @bnchrch changes are in


cpdeethree:
> Approving, but perhaps hold off on merge until masked data interceptor refactor is in?
> 
> That PR removes any reliance on yaml files in the gradle tasks

Sounds good. We can update this branch with those changes as soon as they go in (I've already approved them, but would like Evan or at least one other to approve as well first)

PR #15, opened 2023-02-17
Refactor MaskedDataInterceptor to generate from catalog

bnchrch:
## What
`MaskedDataInterceptor` depends on `specs_secrets_mask.yaml` which is generated by the raw spec files. This will fail in the new monorepo

## How
This changes this by generating the `specs_secrets_mask.yaml` from the catalog.

## Notes for reviewers
1. Ideally I wanted this to use a implementation of `DefinitionProvider` however the gradle dependency graph prevented that due to a circular dependency. Is there a better submodule higher in the tree I can move the compilation of the `DefinitionProvider`'s to so that I can use the implementation?
2. I ended up moving the catalog constants to `common` is that correct? this was so that the same path constant was available in the `MaskedDataInterceptor` and `SpecMaskPropertyGenerator`
3. This also deletes tasks related to generating the oss catalog no longer needed in this project


evantahler (commented):
I'm probably not a good person to review this - you have exceded my gradle knowledge.  I'm tagging @colesnodgrass from the platform team who may be more helpful

colesnodgrass:
👍🏻 I'll be able to take a look on Tuesday morning.

cpdeethree:
@colesnodgrass i might request a retroactive review since this one of the things that would block code unfreeze us early. Let me know what you think :)

cpdeethree (approved):
Nice work.

PR #16, opened 2023-02-17
update remote catalog url to be not gcs anymore

cpdeethree:
## What
Use an API to make this more resilient to the underlying technology serving it

## How
This now points to our ELB fronting a "backend bucket" in GCS

### DIFF
--- a/airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
+++ b/airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
@@ -15,7 +15,7 @@ public class CatalogDefinitionsConfig {
       SEED_SUBDIRECTORY + LOCAL_CONNECTOR_CATALOG_FILE_NAME;
 
   private static final String REMOTE_OSS_CATALOG_URL =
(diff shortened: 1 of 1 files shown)


airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
```diff
@@ -15,7 +15,7 @@ public class CatalogDefinitionsConfig {
       SEED_SUBDIRECTORY + LOCAL_CONNECTOR_CATALOG_FILE_NAME;
 
   private static final String REMOTE_OSS_CATALOG_URL =
-      "https://storage.googleapis.com/prod-airbyte-cloud-connector-metadata-service/oss_catalog.json";
+      "https://connectors.airbyte.com/api/v0/catalog/oss_catalog.json";
```
bnchrch, line 18: Im thinking we change this to the url that maps directly to the path.

While I like the API path, I think we can start using that at the time were uploading versions of these catalogs.

```suggestion
      "https://connectors.airbyte.com/oss_catalog.json";
```

cpdeethree: Either works! It's more for whatever backwards compatability you guys want to create for yourselves.

PR #22, opened 2023-02-18
pull docs from connectors repo for fe build

cpdeethree:
## What
This is a temporary solution to enable building docs into the frontend artifact from the connectors repo until connector metadata service is live.

## How
We clone connectors repo HEAD and wholesale copy the entire docs folder over to be consumed by frontend build. This is basically identical to how it worked in the monorepo from the end artifact perspective - though if someone wants to test and work with this behavior locally it forces them to copy over the relevant docs directly - so it is a bit of a DevEx regression.

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
We get docs for end users in the platform after this :)

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -169,6 +169,7 @@ jobs:
           cd /tmp
           git clone --depth 1 https://.:${{ secrets.GITHUB_TOKEN }}@github.com/airbytehq/airbyte 
           cp -R airbyte/docs $dir/docs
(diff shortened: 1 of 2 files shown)


cpdeethree:
I just realized that this step should actually go into publish artifacts tasks instead because these builds don't actually publish anything.

cgardens (approved):

cgardens:
legit. i think this is the right trade off for now.

PR #23, opened 2023-02-18
remove more-secrets env

cpdeethree:
## What
This is no longer necessary because we are using org level secrets

## How
removes the environment from gha files

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌


cgardens:
thank you!!! was so tired of still seeing that.

PR #30, opened 2023-02-20
🪟 🔧 Adjust doc middleware for new repo structure

timroes:
## What

Since the connector docs are no longer part of `airbyte-platform`, but only of `airbytehq/airbyte`, we can't always load them locally during development of the frontend anymore.

This adjusts our middleware so that:

* if you don't have the `airbytehq/airbyte` repository checked out in parallel, the docs are loaded from GitHub directly.
* If you have the repository checked out in parallel, docs will load from the parallel checked out repository locally, so you can make edits and preview it directly.

You'll see a message on the command line indicating which mode you're running in:

screenshot-20230220-112331

screenshot-20230220-112352


flash1293 (approved):
Tested and both ways work fine, LGTM

cgardens (approved):

timroes:
Merging outside of daily deploy, since it only touches development setup.

PR #44, opened 2023-02-21
Create builder projects table

flash1293:
**Port from airbyte repo, already approved in https://github.com/airbytehq/airbyte/pull/22879**

Fixes https://github.com/airbytehq/airbyte/issues/22444

By introducing the `connector_builder_project` table and adding methods to read/write/delete projects to the `ConfigRepository` class, including tests for these methodes.

Not sure about the version for the platform - is this worth a minor version bump or should it be patch?

## How to test

* Run the tests introduced as part of the PR
* Build platform and start via docker, then log into internal database via `docker exec -ti airbyte-db psql -U docker -d airbyte` and inspect created table.

*If unsure, leave it blank.*
- [ ] YES 💚
- [x] NO ❌

## 🚨 User Impact 🚨

No impact


flash1293:
Port from `airbyte` repo, already approved in https://github.com/airbytehq/airbyte/pull/22879

cc @maxi297 @jdpgrailsdev @girarda @xiaohansong

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1600,6 +1602,74 @@ public boolean getConnectionHasAlphaOrBetaConnector(final UUID connectionId) thr
...
   }
 
+  public Optional<ConnectorBuilderProject> getConnectorBuilderProject(final UUID builderProjectId, final UUID workspaceId) throws IOException {
+    return database.query(ctx -> ctx.select(CONNECTOR_BUILDER_PROJECT.asterisk())
+        .from(CONNECTOR_BUILDER_PROJECT)
+        .where(getBuilderProjectIdCondition(builderProjectId, workspaceId))
+        .fetch())
+        .map(DbConverter::buildConnectorBuilderProject)
+        .stream()
+        .findFirst();
```
maxi297, line 1612: If we only care about the first one (and hopefully only one), would this be better in terms of DB performance to have a `limit(1)` somewhere?

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1600,6 +1602,74 @@ public boolean getConnectionHasAlphaOrBetaConnector(final UUID connectionId) thr
...
+  }
+
+  public Stream<ConnectorBuilderProject> getConnectorBuilderProjectsByWorkspace(final UUID workspaceId) throws IOException {
+    final Condition matchByIds = CONNECTOR_BUILDER_PROJECT.WORKSPACE_ID.eq(workspaceId);
+
+    return database
+        .query(ctx -> ctx
+            .select(CONNECTOR_BUILDER_PROJECT.ID, CONNECTOR_BUILDER_PROJECT.WORKSPACE_ID, CONNECTOR_BUILDER_PROJECT.NAME,
```
maxi297, line 1620: I don't see where `getConnectorBuilderProjectsByWorkspace` is used but is this clear to the user that `ConnectorBuilderProject` can be partial?

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1600,6 +1602,74 @@ public boolean getConnectionHasAlphaOrBetaConnector(final UUID connectionId) thr
...
+  public boolean deleteBuilderProject(final UUID builderProjectId, final UUID workspaceId) throws IOException {
+    return database.transaction(ctx -> ctx.deleteFrom(CONNECTOR_BUILDER_PROJECT))
+        .where(getBuilderProjectIdCondition(builderProjectId, workspaceId))
+        .execute() > 0;
+  }
+
+  public void writeBuilderProject(final ConnectorBuilderProject builderProject) throws IOException {
+    database.transaction(ctx -> {
+      final OffsetDateTime timestamp = OffsetDateTime.now();
+      final Condition matchId = CONNECTOR_BUILDER_PROJECT.ID.eq(builderProject.getBuilderProjectId());
```
maxi297, line 1638: Should we use `getBuilderProjectIdCondition` here? If feels like there could be many `ConnectorBuilderProject` with the same project ID and different workspace based on how we delete [here](https://github.com/airbytehq/airbyte-platform/pull/44/files#diff-2d29b293778e290157be96e22294c2df1fdfa7c72a48961d03396776446deeddR1629)

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1600,6 +1602,74 @@ public boolean getConnectionHasAlphaOrBetaConnector(final UUID connectionId) thr
...
+      final boolean isExistingConfig = ctx.fetchExists(select()
+          .from(CONNECTOR_BUILDER_PROJECT)
+          .where(matchId));
+
+      if (isExistingConfig) {
```
maxi297, line 1643: I feel wary of duplication here. Should we do something like:

```
      <whatever type> databaseOperation = isExistingConfig ? ctx.update(CONNECTOR_BUILDER_PROJECT) : ctx.insertInto(CONNECTOR_BUILDER_PROJECT).set(WORKSPACE.CREATED_AT, timestamp)
      
      databaseOperation.set(CONNECTOR_BUILDER_PROJECT.ID, builderProject.getBuilderProjectId())
            .set(CONNECTOR_BUILDER_PROJECT.WORKSPACE_ID, builderProject.getWorkspaceId())
            .set(CONNECTOR_BUILDER_PROJECT.NAME, builderProject.getName())
            .set(CONNECTOR_BUILDER_PROJECT.ACTOR_DEFINITION_ID, builderProject.getActorDefinitionId())
            .set(CONNECTOR_BUILDER_PROJECT.MANIFEST_DRAFT, JSONB.valueOf(Jsons.serialize(builderProject.getManifestDraft())))
            .set(WORKSPACE.UPDATED_AT, timestamp)
            .execute();
```

I'm not sure if this is possible because maybe `ctx.update(CONNECTOR_BUILDER_PROJECT)` and `ctx.insertInto(CONNECTOR_BUILDER_PROJECT).set(WORKSPACE.CREATED_AT, timestamp)` do not return the same type but this would ensure that the "core" write operations are shared for the create and the update

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1600,6 +1602,74 @@ public boolean getConnectionHasAlphaOrBetaConnector(final UUID connectionId) thr
...
+            .set(CONNECTOR_BUILDER_PROJECT.ID, builderProject.getBuilderProjectId())
+            .set(CONNECTOR_BUILDER_PROJECT.WORKSPACE_ID, builderProject.getWorkspaceId())
+            .set(CONNECTOR_BUILDER_PROJECT.NAME, builderProject.getName())
+            .set(CONNECTOR_BUILDER_PROJECT.ACTOR_DEFINITION_ID, builderProject.getActorDefinitionId())
+            .set(CONNECTOR_BUILDER_PROJECT.MANIFEST_DRAFT, JSONB.valueOf(Jsons.serialize(builderProject.getManifestDraft())))
+            .set(WORKSPACE.CREATED_AT, timestamp)
+            .set(WORKSPACE.UPDATED_AT, timestamp)
+            .execute();
+      }
+      return null;
```
maxi297, line 1664: Why the return statement here?

maxi297 (commented):
I would prefer if someone else approve since I don't know much about this part of the code but functionally, it seems fine. Just a couple of questions for my understanding

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: There can't be more than one because the project id is the PK

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: It's used in a follow-up PR - it should be clear from the context, having a separate type for that (`ConnectorBuilderProject` vs `ConnectorBuilderProjectWithManifest`) seemed over-engineered to me, but no strong opinion.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: There should only be one row per project id as it's the PK - I changed how to delete in the follow-up PR here: https://github.com/airbytehq/airbyte-platform/pull/45

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: Yeah, not too sure about this. I followed the pattern established in the file as the consistency seemed to outweigh saving a few lines, but also no strong opinion on this.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: It's a type-inference thing - without the explicit `null` return it doesn't have a type param for the generic type of the wrapping db context helper.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1600,6 +1602,74 @@ public boolean getConnectionHasAlphaOrBetaConnector(final UUID connectionId) thr
...
+                CONNECTOR_BUILDER_PROJECT.ACTOR_DEFINITION_ID)
+            .from(CONNECTOR_BUILDER_PROJECT)
+            .where(matchByIds)
+            .fetch())
+        .map(DbConverter::buildConnectorBuilderProjectWithoutManifestDraft)
+        .stream();
+  }
+
+  public boolean deleteBuilderProject(final UUID builderProjectId, final UUID workspaceId) throws IOException {
+    return database.transaction(ctx -> ctx.deleteFrom(CONNECTOR_BUILDER_PROJECT))
```
xiaohansong, line 1630: A thought here - should we actually delete this from DB? It won't leave any record to debug. Normal practice would set a tombstone flag here and avoid querying this when listing/getting builder project. Though for GDPR reasons we need to remember to delete this record if requested.

wdyt?

xiaohansong (commented):
Have one comment about deletion, otherwise code looks great!

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: I considered this when setting up the tech spec but ultimately decided against it because there is no strong technical reason that requires it (like other tables referencing it and we don't want to cascade deletes or we need to keep it around for jobs that are still running or something like that).

But if it's unexpected in this case I think that makes it worth implementing it with a soft delete from the start - it will probably be handy at some point in the future. I will adjust the PR.

flash1293:
@xiaohansong switched to soft delete, could you take another look?

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
girarda: I personally don't mind the duplication because it improves the readability of the SQL command.

girarda (approved):

flash1293:
@xiaohansong marking this as ready-to-ship as Alex approved here and you approved over in the airbyte repo already.

xiaohansong (approved):

octavia-approvington:
Myomoto says it looks good <br/> thats a niiice

octavia-approvington (approved):

PR #45, opened 2023-02-21
Connector builder apis

flash1293:
Based on https://github.com/airbytehq/airbyte-platform/pull/44

Fixes https://github.com/airbytehq/airbyte/issues/22570
Fixes https://github.com/airbytehq/airbyte/issues/22571
Fixes https://github.com/airbytehq/airbyte/issues/22572
Fixes https://github.com/airbytehq/airbyte/issues/22445
Fixes https://github.com/airbytehq/airbyte/issues/22573

This PR is adding the API layer for creating, updating, deleting, listing and retrieving builder projects based on the new model. This includes the following parts:
* Extend openapi spec for the server (`airbyte-api/src/main/openapi/config.yaml`)
* Implement API controller (`airbyte-server/src/main/java/io/airbyte/server/apis/ConnectorBuilderProjectApiController.java`) which is forwarding everything to the handler
* Implement handler (`airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java`) which is executing the write operations with help of the `ConfigRepository`
  * Checking whether the project exists in the current workspace is centralized in a `validateWorkspace` method that check with the database
* Add tests for the handler
* Extend open api spec: `airbyte-api/src/main/openapi/config.yaml`
* Extend the builder project fetching with a calculated field `hasDraft` (`airbyte-config/config-models/src/main/resources/types/ConnectorBuilderProject.yaml`, `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/DbConverter.java`, `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java`) which is used for the listing to know whether there is a draft without actually fetching it.
* Fix an existing problem of not storing a string `"null"` instead of actually no value if there is no manifest draft defined (also adding tests for this) - this happened because JSON serialization will turn a Java `null` into a JSON string `"null"`. Added a branch to https://github.com/airbytehq/airbyte/pull/23143/files#diff-2d29b293778e290157be96e22294c2df1fdfa7c72a48961d03396776446deeddR1663 to fix this

Notes:
* I tried to follow established approaches whereever possible to make this code as little "suprising" as possible for a platform engineer, but maybe I misunderstood something
* In theory it would be possible in some places to collapse multiple queries into one (e.g. the validate call is already implicitly checking whether a builder project exists, but the write call is doing another lightweight exists call), however separating different concerns seemed more important here. If it turns out to be necessary, performance optimizations can be done in a later step.

Can be tested end-to-end by building the server via `SUB_BUILD=PLATFORM ./gradlew assemble`, starting via `BASIC_AUTH_USERNAME="" BASIC_AUTH_PASSWORD="" VERSION=dev docker-compose --file ./docker-compose.yaml up` and using curl to call the new endpoints while checking the state of the database by logging into the database via `docker exec -ti airbyte-db psql -U docker -d airbyte`

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨

No impact


flash1293:
Port from `airbyte` repo, already approved by @xiaohansong https://github.com/airbytehq/airbyte/pull/23063

cc @maxi297 @jdpgrailsdev @girarda @xiaohansong

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -2502,6 +2551,53 @@ components:
...
+          type: object
+          additionalProperties: true
+    ConnectorBuilderProjectIdWithWorkspaceId:
+      type: object
+      required:
+        - workspaceId
+        - builderProjectId
+      properties:
+        workspaceId:
+          $ref: "#/components/schemas/WorkspaceId"
```
maxi297, line 2575: Using this, it means that we are binding with the definition of a "workspaceId" in the platform. This means that if platform wants to change that, we will be impacted. Is this something that makes sense to us?

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandlerTest.java
```diff
@@ -0,0 +1,150 @@
...
+    when(configRepository.getConnectorBuilderProject(project.getBuilderProjectId(), false)).thenReturn(project);
+
+    final ExistingConnectorBuilderProjectWithWorkspaceId update = new ExistingConnectorBuilderProjectWithWorkspaceId()
+        .builderProject(new ConnectorBuilderProjectDetails().name(project.getName())
+            .draftManifest(new ObjectMapper().convertValue(project.getManifestDraft(), new TypeReference<Map<String, Object>>() {})))
+        .workspaceId(workspaceId).builderProjectId(project.getBuilderProjectId());
+
+    connectorBuilderProjectsHandler.updateConnectorBuilderProject(update);
+
+    verify(uuidSupplier, never()).get();
```
maxi297, line 94: Why is this important for this test? If feels like it's not a relevant outcome if it's called or not

flash1293: Fair point, I'm going to remove this.

airbyte-api/src/main/openapi/config.yaml
flash1293: The `WorkspaceId` schema is just a uuid-string at the moment to make sure this compiles to a `UUID` member, but in general I think it makes sense to get an error if this assumption changes - after all we also have a `workspaceId` column in the builder project model.

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandlerTest.java
```diff
@@ -0,0 +1,150 @@
...
+    verify(configRepository, atMostOnce())
```
xiaohansong, line 95: nit: should this be times(1)?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    return new ConnectorBuilderProject().withBuilderProjectId(id).withWorkspaceId(projectCreate.getWorkspaceId())
+        .withName(projectCreate.getBuilderProject().getName())
+        .withManifestDraft(new ObjectMapper().valueToTree(projectCreate.getBuilderProject().getDraftManifest()));
+  }
+
+  private ConnectorBuilderProjectIdWithWorkspaceId idResponseFromBuilderProject(final ConnectorBuilderProject project) {
+    return new ConnectorBuilderProjectIdWithWorkspaceId().workspaceId(project.getWorkspaceId()).builderProjectId(project.getBuilderProjectId());
+  }
+
+  public void validateWorkspace(final UUID projectId, final UUID workspaceId) throws ConfigNotFoundException, IOException {
```
xiaohansong, line 61: should this be private?

xiaohansong (approved):
two minor comments, otherwise lgtm!

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -2502,6 +2551,53 @@ components:
...
+    ConnectorBuilderProjectWithWorkspaceId:
+      type: object
+      required:
+        - workspaceId
+        - builderProject
+      properties:
+        workspaceId:
+          $ref: "#/components/schemas/WorkspaceId"
+        builderProject:
+          $ref: "#/components/schemas/ConnectorBuilderProjectDetails"
```
lmossman, line 2600: @flash1293 I don't think this should be pointing to `ConnectorBuilderProjectDetails`, because that struct has `builderProjectId` as a required field. 

This `ConnectorBuilderProjectWithWorkspaceId` struct is used for the request body `/connector_builder_projects/create`, so when that is called the webapp will not know what the `builderProjectId` should be, however with the current API setup this field is required when calling it.

So I think we'll need a new struct similar to `ConnectorBuilderProjectDetails` but with `builderProjectId` not being required, to be used here

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        .withName(projectCreate.getBuilderProject().getName())
+        .withManifestDraft(new ObjectMapper().valueToTree(projectCreate.getBuilderProject().getDraftManifest()));
+  }
+
+  private ConnectorBuilderProject builderProjectFromCreate(final ConnectorBuilderProjectWithWorkspaceId projectCreate) {
+    final UUID id = uuidSupplier.get();
+
+    return new ConnectorBuilderProject().withBuilderProjectId(id).withWorkspaceId(projectCreate.getWorkspaceId())
+        .withName(projectCreate.getBuilderProject().getName())
+        .withManifestDraft(new ObjectMapper().valueToTree(projectCreate.getBuilderProject().getDraftManifest()));
```
lmossman, line 53: Small note: this currently isn't respecting the `hasDraft` value of the `ConnectorBuilderProjectWithWorkspaceId` object that is passed into this function.

I tried calling this with `hasDraft` set to `false` and let `draftManifest` unset, but when I then called the `/list` endpoint, it returned `hasDraft: true` for this project, because this function is always setting the `manifestDraft` so something regardless of the `hasDraft` value.

Thinking about this more, the request body for this `/create` API endpoint probably shouldn't take in a `hasDraft` field at all, since that just meant to be an informative field when listing. So related to my above comment, I think `hasDraft` and `builderProjectId` should be removed from the `/create` request body.

And this method should be updated to check if the `draftManifest` is null on the `projectCreate` argument, and if so it should not set it on the database record.

airbyte-api/src/main/openapi/config.yaml
flash1293: You are right, we need different classes for the request and the response. However, as `ConnectorBuilderProjectDetails` is only used for writing on this PR I will fix this issue on https://github.com/airbytehq/airbyte-platform/pull/46

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
flash1293: You are right, `hasDraft` is only relevant for the response, not for the request. I will separate these two cases in the other PR https://github.com/airbytehq/airbyte-platform/pull/46 (it also fixes the issue with `draftManifest` always being set to something even if you unset it)

maxi297 (approved):
LGTM!

lmossman (approved):
LGTM given that the issues I called out have been addressed in https://github.com/airbytehq/airbyte-platform/pull/46

girarda (approved):

flash1293:
I discovered one small issue - when returning a manifest, the serialization logic would omit JSON structures which don't contain a primitive value (string, number, boolean), for example in `{"a": 213, "b": {"c":[]}}`, the branch starting from "b" would be missing from the output.

This could be fixed by changing the type of the api from
```
type: object
additionalProperties: true
```

to unspecified type (this is also how the same problem was solved in other places in our openapi spec). Adjusted a unit test to cover this.

PR #46, opened 2023-02-21
Connector builder projects: List and get APIs

flash1293:
## What

Based on https://github.com/airbytehq/airbyte-platform/pull/45

Fixes https://github.com/airbytehq/airbyte/issues/22445
Fixes https://github.com/airbytehq/airbyte/issues/22573

By adding APIs to list all builder projects of a workspace as well as fetching a single project with its manifest.

## How

* Extend open api spec: `airbyte-api/src/main/openapi/config.yaml`
* Extend the builder project fetching with a calculated field `hasDraft` (`airbyte-config/config-models/src/main/resources/types/ConnectorBuilderProject.yaml`, `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/DbConverter.java`, `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java`) which is used for the listing to know whether there is a draft without actually fetching it.
* Fix an existing problem of not storing a string `"null"` instead of actually no value if there is no manifest draft defined (also adding tests for this) - this happened because JSON serialization will turn a Java `null` into a JSON string `"null"`. Added a branch to https://github.com/airbytehq/airbyte/pull/23143/files#diff-2d29b293778e290157be96e22294c2df1fdfa7c72a48961d03396776446deeddR1663 to fix this

For testing and general notes see description of https://github.com/airbytehq/airbyte/pull/23063

## Can this PR be safely reverted / rolled back?
*If you know that your PR is backwards-compatible and can be simply reverted or rolled back, check the YES box.*

*Otherwise if your PR has a breaking change, like a database migration for example, check the NO box.*

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨

No impact

PR #47, opened 2023-02-21
Introduce config injection

flash1293:
Based on https://github.com/airbytehq/airbyte-platform/pull/46

## What

Fixes https://github.com/airbytehq/airbyte/issues/22439
Fixes https://github.com/airbytehq/airbyte/issues/22440

Adds actor definition config injection table that injects a json blob under a specified injection path into the connection configration of source or destination before running a sync/check/discover job.

I slightly deviated from the description in by avoiding a UUID `id` and instead going with a composite PK over injection_path and actor_definition_id. Semantically it doesn't make sense to have more than one config injection that targets the same injection path and actor definition id and when writing them as part of #22457 injection_path and actor_definition_id will be the identifying properties as well - introducing a synthetic UUID will only complicate things in this specific case, especially since these objects are not planned to be exposed via API at any point.

## How

* Add a new table and models as usual.
* Introduce `ConfigInjector` helper class that fetches and injects blobs into the config and make it known to DI for worker and server.
* Call the helper in all the places a job config is put together:
  * `airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/DefaultSynchronousSchedulerClient.java`
  * `airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/factory/DefaultSyncJobFactory.java`
  * `airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java`
* The write method is not used yet but will be used in a follow-up PR
* Adds tests for new methods in config repository as well as config injector
  * I only introduced a single test suite for both parts together because they are fairly slim already and I don't see how we gain much by having separate tests
* Extends existing tests for the places where job configs are put together

## How to test

To test this end-to-end, set up a connection with the pokeapi source, then execute the following query:
```
insert into actor_definition_config_injection values ('"pikachu"', 'pokemon_name', '6371b14b-bc68-4236-bfbd-468e8df8e968');
```

This will cause all sync/discover/check jobs to run for pikachu, no matter what's actually configured. Setting to an invalid value should fail the calls with the expected error message.

*If unsure, leave it blank.*
- [ ] YES 💚
- [x] NO ❌

## 🚨 User Impact 🚨

No impact


flash1293:
@xiaohansong @jdpgrailsdev This PR is ready for review - especially interested in feedback about the parts touching the places in the code where config is actually injected in the worker and server.

airbyte-config/config-models/src/main/resources/types/ActorDefinitionConfigInjection.yaml
```diff
@@ -0,0 +1,21 @@
...
+description: JSON object that is injected into every run of the referenced actor definition
+type: object
+required:
+  - actorDefinitionId
+  - jsonToInject
+  - injectionPath
+additionalProperties: false
+properties:
+  injectionPath:
+    type: string
```
girarda, line 14: nit: can you add a description or an example for future reference?

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigInjector.java
```diff
@@ -0,0 +1,34 @@
...
+public class ConfigInjector {
+
+  private final ConfigRepository configRepository;
+
+  public ConfigInjector(final ConfigRepository configRepository) {
+    this.configRepository = configRepository;
+  }
+
+  public JsonNode injectConfig(final JsonNode configuration, final UUID actorDefinitionId) throws IOException {
+    final List<ActorDefinitionConfigInjection> configInjections = configRepository.getActorDefinitionConfigInjections(actorDefinitionId).toList();
```
girarda, line 27: nit: is there a reason to turn the stream into a list just to apply a `forEach`? You should be apply to apply the `forEach` directly on the stream

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_00_002__AddActorDefinitionConfigInjection.java
```diff
@@ -0,0 +1,57 @@
...
+    // As database schema changes, the generated jOOQ code can be deprecated. So
+    // old migration may not compile if there is any generated code.
+    try (final DSLContext ctx = DSL.using(context.getConnection())) {
+      addConfigInjectionTable(ctx);
+    }
+  }
+
+  private static void addConfigInjectionTable(final DSLContext ctx) {
+    final Field<JSONB> jsonToInject = DSL.field("json_to_inject", SQLDataType.JSONB.nullable(false));
+    final Field<String> injectionPath = DSL.field("injection_path", SQLDataType.VARCHAR(256).nullable(false));
```
girarda, line 43: is there a reason to use `VARCHAR(256)` instead of `TEXT`? As far as I know, there's no benefit to using a VARCHAR on postgres.

https://stackoverflow.com/a/4849030

girarda (approved):
🚂

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_00_002__AddActorDefinitionConfigInjection.java
flash1293: jooq has the following string data types: VARCHAR, CHAR, LONGVARCHAR, CLOB, NVARCHAR, NCHAR, LONGNVARCHAR, NCLOB

`VARCHAR` with a length should be converted to a `VARCHAR` postgres type while `VARCHAR` without length will probably become `TEXT` (both are used in other tables as well)

Happy to go without a length limitation here, as you say there is no good reason to restrict it.

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_01_001__AddActorDefinitionConfigInjection.java
```diff
@@ -0,0 +1,57 @@
...
+import org.flywaydb.core.api.migration.Context;
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.JSONB;
+import org.jooq.impl.DSL;
+import org.jooq.impl.SQLDataType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// TODO: update migration description in the class name
```
maxi297, line 24: Is this a TODO that needs to be done as part of this PR? If not, why can't it be done right now?

flash1293:
@xiaohansong @jdpgrailsdev PR is ready for review, there is one test failure though I'm not sure what to do about.

The build failed with
```
org.mockito.exceptions.misusing.UnnecessaryStubbingException: |  
  | Unnecessary stubbings detected. |  
  | Clean & maintainable test code requires zero unnecessary code. |  
  | Following stubbings are unnecessary (click to navigate to relevant line of code): |  
  | 1. -> at io.airbyte.workers.temporal.scheduling.activities.GenerateInputActivityTest.setUp(GenerateInputActivityTest.java:106)
```

but the stub introduced here (`when(configInjector.injectConfig(any(), any())).thenAnswer(i -> i.getArguments()[0]);`) is actually necessary - if I remove this line test fail. Do you know what's going on here?

maxi297 (approved):
LGTM!

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_01_001__AddActorDefinitionConfigInjection.java
flash1293: that's a leftover, removing...

airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/DefaultSynchronousSchedulerClient.java
```diff
@@ -79,14 +83,14 @@ public SynchronousResponse<StandardCheckConnectionOutput> createSourceCheckConne
     final JobCheckConnectionConfig jobCheckConnectionConfig = new JobCheckConnectionConfig()
         .withActorType(ActorType.SOURCE)
         .withActorId(source.getSourceId())
-        .withConnectionConfiguration(sourceConfiguration)
+        .withConnectionConfiguration(configInjector.injectConfig(sourceConfiguration, source.getSourceDefinitionId()))
```
maxi297, line 86: When we call `withConnectionConfiguration`, will we always want to have the config injection? If the config is always the same and there will always be injection, should we wrap this in a "ConnectionConfigurationDAO"? Else, we will need to modify everywhere the config is used every time we change something associated to it

flash1293: Something we could do right now that might be a good preparation is to make the `ConfigInjector` also take care of calling the `OAuthConfigSupplier` as they have similar parameters and I think at the moment both of them are called in the same places. OTOH it might be more confusing than helpful to tie these things together right now.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigInjector.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.micronaut.context.annotation.Requires;
+import jakarta.inject.Singleton;
+import java.io.IOException;
+import java.util.UUID;
+
+@Singleton
+@Requires(bean = ConfigRepository.class)
```
jdpgrailsdev, line 15: As discussed, this will only create this singleton when the `ConfigRepository` is present as a singleton.  Because of this, you may want to inject `Optional<ConfigInjector>` everywhere it is used to avoid a possible null pointer.  You can simply modify the constructor parameter to take `Optional<ConfigInjector>` and Micronaut will do the rest.

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_01_001__AddActorDefinitionConfigInjection.java
```diff
@@ -0,0 +1,56 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(V0_41_01_001__AddActorDefinitionConfigInjection.class);
+
+  @Override
+  public void migrate(final Context context) throws Exception {
+    LOGGER.info("Running migration: {}", this.getClass().getSimpleName());
+
+    // Warning: please do not use any jOOQ generated code to write a migration.
+    // As database schema changes, the generated jOOQ code can be deprecated. So
+    // old migration may not compile if there is any generated code.
+    try (final DSLContext ctx = DSL.using(context.getConnection())) {
```
jdpgrailsdev, line 35: Heads up that this is going to conflict with an open PR that I have:  https://github.com/airbytehq/airbyte-platform/pull/128/.  We are attempting to update to a modern version of jOOQ (we are two years behind in terms of releases).  One of the changes is that the `DSLContext` class is no longer auto-closeable, so if my PR lands first, you will simply need to remove the try-with-resource block.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigInjector.java
jdpgrailsdev: It is unavailable in the data-plane version of airbyte-worker.  In that case, we don't want to fail startup because we cannot create the ConfigInjector.

flash1293: Discussed this in a call, for now the current approach makes sense as cloud acceptance tests will catch the problem if someone tries to use this bean from the date plane.

Once the platform team tackles the project of removing all direct data base access from workers, an API will have to be introduced for it.

jdpgrailsdev (approved):
:shipit: As @flash1293 mentioned in the comments, we discussed the control-plane vs data-plane impact and decided that the singleton in question is only being used in the workers in code that already depends on the `ConfigRepository` so this is not adding any additional risk to the data-plane configuration.  The ultimate solution is to remove all database access from the workers to eliminate this concern entirely.

xiaohansong (approved):

xiaohansong:
Looks like this is getting called before dispatched to individual task queue - so that's totally on control plane and it's a valid use case.

PR #48, opened 2023-02-21
🪟🐛 Prevent invalid entries for array of numbers

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/22850

## How
https://github.com/airbytehq/airbyte/issues/22850 is caused by the fact that the `TagInput` control doesn't validate whether the typed input is confirming to the type specified in the schema. In case of number or integer entries in an array this leads to errors during casting the form values which is done in multiple places.

This PR fixes the problem using a `type="number"` input instead so it's not even possible to enter invalid numbers in the first place.

Note: This implementation does not allow to enter multiple values using the `,` or `;` key to enter multiple values (Enter or Tab have to be used), didn't find a good solution for this, correctness seemed more important to me than ease-of-use features and given that it's a rare case (this is the only place a numeric array is used in existing connectors AFAICT), I didn't want to go down this rabbit hole right now.


airbyte-webapp/src/core/form/schemaToFormBlock.ts
```diff
@@ -119,6 +119,15 @@ export const jsonSchemaToFormBlock = (
     };
   }
 
+  const type = (Array.isArray(jsonSchema.type) ? jsonSchema.type[0] : jsonSchema.type) ?? "null";
```
lmossman, line 122: It looks like here we are just grabbing the first value of `type` if it is an array, but on line 128 we reject `items` if its type is an array. 

Is there a reason the two shouldn't be consistent? If no then maybe we should grab the first value from `jsonSchema.items.type` if it is an array as well?

airbyte-webapp/src/components/ui/TagInput/TagInput.tsx
```diff
@@ -1,14 +1,19 @@
...
 import styles from "./TagInput.module.scss";
 
-const components = {
+const overwrittenComponents = {
   DropdownIndicator: null,
 };
 
+const NumberInput: ComponentType<InputProps<Tag, true, GroupBase<Tag>>> = (
+  props: InputProps<Tag, true, GroupBase<Tag>>
+) => <components.Input {...props} type="number" className={classNames(props.className, styles.hideArrowButtons)} />;
```
lmossman, line 15: I found an issue when testing this locally: if you type an `e` into the input, it accepts it (I believe this is expected because `e` can be used to specify a number in scientific notation), but the character is halfway cut off in the input. I can type more numbers after the `e`, but they are still cut off, and I cannot erase them by pressing backspace. And if I try to hit enter, it shows me the default `Please enter a number` warning. 

Here is a loom to demonstrate: https://www.loom.com/share/9a0429912ed049968072cc3b760f9dfc

I think allowing `e` here is probably fine since that can technically still be used for a number, but the visual issue should probably try to be fixed? If it ends up being really difficult to fix then we can always do that in a separate PR

lmossman (approved):
Had a couple of comments and found one visual bug which we may want to try to address before merging

airbyte-webapp/src/core/form/schemaToFormBlock.ts
flash1293: The fact that we grab the first one for the top level type is a leftover because we currently have a few specs which set the type like this: `type: ["string", "null"]`. I would like to get rid of that eventually, but it requires a bunch of cleanup first.

For arrays we don't allow this by the acceptance test criteria:

Special case for type + null:
https://github.com/airbytehq/airbyte/blob/d82e01a34fe5d637fbdda394f5cad4ab11a78ffb/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py#L290

Always fail if this is done in arrays:
 https://github.com/airbytehq/airbyte/blob/d82e01a34fe5d637fbdda394f5cad4ab11a78ffb/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py#L336

airbyte-webapp/src/components/ui/TagInput/TagInput.tsx
flash1293: Good call, this happens because a number input will set its `value` to `''` if it can't be parsed as a number which breaks react-select because it checks the `value` to layout it.

I changed the approach completely by parsing the input as a number before it gets put into the state and filtering out invalid entries. It kind of works, but it's not perfect either because it could be confusing if the user enters something, then hits Enter and it's suddenly gone again because it couldn't be parsed. On the plus side this approach still keeps `,`/`;` separated lists of numbers working.

It's a really small thing, but super annoying to solve. What do you think? Pretty much out of better ideas... The only other solution would be to drop this feature completely and only allow string inputs which then have to be parsed connector side (we can still do client side validation using a pattern).

airbyte-webapp/src/core/form/schemaToFormBlock.ts
lmossman: Ah that makes sense, I think this logic is good then

airbyte-webapp/src/components/ui/TagInput/TagInput.tsx
lmossman: I think this solution is good, I like that is brings back the `,`/`;` functionality so it is more consistent with other types of arrays. 

I agree it is a little bit confusing that if you type in a string without a number and hit enter, it just disappears. But this is still a huge improvement on the page crashing and contains no weird visual bugs, so I think we should go with it.

To make this less confusing in the future, maybe as part of the field format changes in the Simplify connector form project, we should also show an indication of which type of values should go into these array inputs to make it clear to users that only numbers can be entered in this case.

lmossman (approved):
LGTM

airbyte-webapp/src/components/ui/TagInput/TagInput.tsx
flash1293: > To make this less confusing in the future, maybe as part of the field format changes in the Simplify connector form project, we should also show an indication of which type of values should go into these array inputs to make it clear to users that only numbers can be entered in this case.

I like that a lot, great idea.

octavia-approvington:
This code is at peace with itself <br/> imagine karate kid approving

octavia-approvington (approved):

PR #49, opened 2023-02-21
🪟🐛 Connector builder: Replace instead of navigate

flash1293:
This PR fixes the issue that navigating back after going to the builder UI currently doesn't work.

The problem is that the user is redirected from the landing page to the edit page, but this navigation is writing a new history entry, so navigating "back" will lead the user to the landing page again, which will immediately redirect them again.

This PR sets the replace option which replaces the landing page history entry instead of writing a new one.

## How to test

* Enable the builder navigation links experiment (`.experiments.json`):
```
{
    "connectorBuilder.showNavigationLinks": true
}
```
* Click the "Builder" nav item
* Click back button
* You get back to where you came from


lmossman (approved):
The explanation and the change make sense to me, tested locally and it fixed the issue. Thanks!

octavia-approvington:
What are we doing again? <br/> merge and squash

octavia-approvington (approved):

PR #50, opened 2023-02-21
🪟🔧 Add testid for TagInput input element

flash1293:
Depends on https://github.com/airbytehq/airbyte-platform/pull/48

This PR adds an explicit test id to the tag input component based on the name to find it in E2E tests. To do so, an overwritten `Input` component is passed in which is setting the `data-testid`:
Screenshot 2023-02-21 at 11 53 52


lmossman (approved):
LGTM. Did you make this change because we want to access the Input in an E2E test right now, or just in case we want to in the future?

flash1293:
It came up in the context of the E2E tests @maxi297 is writing right now.

PR #52, opened 2023-02-21
Fix #22687

xpuska513:
## What
* fix for missing configmap values(fixes https://github.com/airbytehq/airbyte/issues/22687)


cpdeethree (approved):
I think it makes sense to disable caching for these tests

PR #54, opened 2023-02-21
🪟 🔧 Enable local docs for airbyte-platform-internal

timroes:
## What

This PR ensures that serving local docs will work no matter if you're running from `airbyte-platform` or `airbyte-platform-internal` (which will be the case next week). In `airbyte-platform-internal` the OSS code will be nested under `/oss` and thus we need to look one folder higher to find the parallel checked out `airbyte` folder for local serving docs.

This PR tries both pathes now. We decided for the simple path checking solution over actually using git to retrieve the root command, to also work in case users would download the files via zip and not have a git repository instantiated.

### DIFF
--- a/airbyte-webapp/packages/vite-plugins/doc-middleware.ts
+++ b/airbyte-webapp/packages/vite-plugins/doc-middleware.ts
@@ -6,25 +6,22 @@ import path from "path";
 import chalk from "chalk";
 import express from "express";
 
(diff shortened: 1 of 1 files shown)


flash1293 (approved):
Tested all three cases locally and all of them work as expected:
* connectors repo next to platform repo
* no connectors repo
* connectors repo one level up from platform repo

timroes:
Merging outside cycle, since it's only touching development tooling and no production code.

PR #56, opened 2023-02-21
[ISSUE #22954] adding e2e test for connector builder

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte/issues/22954

## How
By adding e2e tests

## Recommended reading order
1. `airbyte-webapp-e2e-tests/cypress/integration/connectorBuilder.spec.ts` are the actual tests
2. `airbyte-webapp-e2e-tests/cypress/commands/connectorBuilder.ts` and `airbyte-webapp-e2e-tests/cypress/pages/connectorBuilderPage.ts` are the methods I needed to change to allow testing
3. I had to add handles to ensure I was able to select the elements in the DOM properly

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
These tests will run in the pipeline when merged


airbyte-webapp-e2e-tests/cypress/commands/connectorBuilder.ts
```diff
@@ -66,4 +91,54 @@ export const assertMultiPageReadItems = () => {
...
+'   "properties": {\n' +
+'     "name": {\n' +
+'       "type": "string"\n' +
+'     }\n' +
+'   },\n' +
+'   "type": "object"\n' +
+' }'
+export const assertSchema = () => {
+  cy.get('[data-testid="tag-tab-detected-schema"]').click();
+  cy.get('pre[class*="SchemaDiffView"]').contains(SCHEMA).should("exist");
```
maxi297, line 130: I feel this asserting is very precise and could break easily. For example, the JSON starts with a white space. If this change, this test will fail even though this might be perfectly fine

airbyte-webapp-e2e-tests/cypress/commands/connectorBuilder.ts
```diff
@@ -66,4 +91,54 @@ export const assertMultiPageReadItems = () => {
...
+const SCHEMA_WITH_MISMATCH = '{{}"$schema": "http://json-schema.org/schema#", "properties": {{}"name": {{}"type": "number"}}, "type": "object"}'
+export const acceptSchemaWithMismatch = () => {
+  cy.get('[data-testid="tag-tab-stream-schema"] button').click();
+  cy.get('textarea').type(SCHEMA_WITH_MISMATCH);
+};
+
+export const assertSchemaMismatch = () => {
+  cy.get('[data-testid="tag-tab-detected-schema"]').click();
+  cy.contains("Detected schema and declared schema are different").should("exist");
+  cy.get('pre[class*="SchemaDiffView"]').contains('-      "type": "number"').should("exist");
```
maxi297, line 142: This might be too precise and lead to test flakiness. If we agree with this, I can keep the `cy.contains("Detected schema and declared schema are different").should("exist");` and remove the rest

airbyte-webapp-e2e-tests/cypress/integration/connectorBuilder.spec.ts
```diff
@@ -1,41 +1,113 @@
...
     configureStream();
   });
 
-  it("Fail on missing auth", () => {
+  afterEach(() => {
+    cleanUp();
+  });
+
+  /*
+  This test assumes it runs before "Read - Without pagination or partition router" since auth will be configured at that
```
maxi297, line 39: This was already the case before and it felt costly to fix so I just added a comment

airbyte-webapp-e2e-tests/cypress/integration/connectorBuilder.spec.ts
```diff
@@ -1,41 +1,113 @@
...
-  it("Succeed on provided auth", () => {
+  it("Read - Without pagination or partition router", () => {
     configureAuth();
     testStream();
     assertTestReadItems();
   });
 
-  it("Pagination", () => {
+  /*
+  All the tests below assume they run after "Read - Without pagination or partition router" in order to have auth
```
maxi297, line 54: This was already the case before and it felt costly to fix so I just added a comment

airbyte-webapp-e2e-tests/cypress/pages/connectorBuilderPage.ts
```diff
@@ -94,3 +123,12 @@ export const testStream = () => {
...
   cy.get(testStreamButton).click();
 };
+
+const GO_BACK_AND_GO_NEXT_BUTTONS = 2
+export const assertHasNumberOfPages = (numberOfPages: number) => {
+  for (var i = 0; i < numberOfPages; i++) {
+    cy.get(testPageItem).contains(i + 1).should('exist')
+  }
+
+  cy.get(testPageItem).should('have.length', numberOfPages + GO_BACK_AND_GO_NEXT_BUTTONS)
```
maxi297, line 133: I feel this assertion is a bit weird but the for loop is not sufficient since there could be more elements that exist. However, if the arrows placement is changed, this test will fail

maxi297:
I had some flakiness on many "type" and `check` methods and ended up adding a lot for `{ force: true }` which feels like a sloppy solution

airbyte-webapp-e2e-tests/dummy_api.js
```diff
@@ -5,19 +5,39 @@
...
-const items = [{ name: "abc" }, { name: "def" }, { name: "xxx" }, { name: "yyy" }];
+const itemsWithoutSlices = [{ name: "abc" }, { name: "def" }, { name: "xxx" }, { name: "yyy" }];
+const itemsExceedingReadLimit = Array.from(Array(100).keys()).map(count => ({ exceedingPageLimit: "subitem" + count }))
+const PAGE_SIZE = 2;
+
+const paginateResults = function(result, offset, page_size) {
+  return [...result].splice(offset, page_size)
+}
+
+const generateResults = function(item_id, count) {
```
flash1293, line 16: nit: `count` argument is not used (handled internally)

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -167,18 +167,22 @@ const StreamControls = ({
...
-        onSelect={() => setSelectedTab("configuration")}
-      />
-      <StreamTab
-        label={formatMessage({ id: "connectorBuilder.streamSchema" })}
-        selected={selectedTab === "schema"}
-        onSelect={() => setSelectedTab("schema")}
-        showErrorIndicator={hasSchemaErrors}
-        showSchemaConflictIndicator={Boolean(formattedDetectedSchema && schema.value !== formattedDetectedSchema)}
-      />
+      <div data-testid="tag-tab-stream-configuration">
```
flash1293: nit - we can move the `data-testid` logic inside the `StreamTab` component, I will push a commit for this.

airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/SliceSelector.tsx
```diff
@@ -60,13 +60,15 @@ export const SliceSelector: React.FC<SliceSelectorProps> = ({ className, slices,
...
   return (
-    <ListBox
-      className={className}
-      options={options}
-      selectedValue={selectedSliceIndex}
-      onSelect={(selected) => onSelect(selected)}
-      buttonClassName={styles.button}
-      controlButton={ControlButton}
-    />
+    <div data-testid="tag-select-slice">
```
flash1293: nit: same here - this can be moved into the `ListBox` component, I will take care of that.

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderField.tsx
```diff
@@ -123,12 +123,14 @@ const InnerBuilderField: React.FC<BuilderFieldProps & FastFieldProps<unknown>> =
         />
       )}
       {props.type === "array" && (
-        <ArrayField
-          name={path}
-          value={(field.value as string[] | undefined) ?? []}
-          setValue={setValue}
-          error={hasError}
-        />
+        <div data-testid={`tag-input-${path}`}>
```
flash1293, line 126: Just a note - I have a PR up here that sets a testid on the actual input: https://github.com/airbytehq/airbyte-platform/pull/50

Can be consumed in a follow-up PR though.

airbyte-webapp-e2e-tests/cypress/pages/connectorBuilderPage.ts
```diff
@@ -94,3 +123,12 @@ export const testStream = () => {
   cy.wait(500);
   cy.get(testStreamButton).click();
 };
+
+const GO_BACK_AND_GO_NEXT_BUTTONS = 2
+export const assertHasNumberOfPages = (numberOfPages: number) => {
+  for (var i = 0; i < numberOfPages; i++) {
```
flash1293, line 129: It seems to me like the best approach for this is to look for the `a` elements instead of the `li`s and don't count `rel="prev"` or `rel="next"` - best way to do it I can see is `cy.get(SELECTOR).not(`[rel="prev"]`).not(`[rel="next"]`).to('have.length'`. I don't think we need to validate the actual contents of those elements.

We can rely on the `rel` attributes here because those are semantic attributes, not just an implementation detail: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel

We still assume a bit about the structure of this component by looking for `a` tags, but I don't see a good way around that and I think that's fairly stable.

airbyte-webapp-e2e-tests/cypress/pages/connectorBuilderPage.ts
flash1293: See comment above, this way you don't need this assertion

airbyte-webapp-e2e-tests/cypress/integration/connectorBuilder.spec.ts
```diff
@@ -1,41 +1,113 @@
...
+  assertSource404Error,
+  cleanUp,
   configureAuth,
   configureGlobals,
-  configureStream,
   configurePagination,
-  assertMultiPageReadItems,
+  configureStream,
+  configureStreamSlicer,
+  invalidAuth,
```
flash1293, line 20: seems like this doesn't exist, is this a leftover?

airbyte-webapp-e2e-tests/cypress/integration/connectorBuilder.spec.ts
flash1293: Good to call it out but this is expected for now to keep the test time low as you mentioned.

airbyte-webapp-e2e-tests/cypress/commands/connectorBuilder.ts
```diff
@@ -66,4 +91,54 @@ export const assertMultiPageReadItems = () => {
...
\ No newline at end of file
+};
+
+const MAX_NUMBER_OF_PAGES = 5;
+export const assertMaxNumberOfPages = () => {
+  assertHasNumberOfPages(MAX_NUMBER_OF_PAGES)
+};
+
+export const assertHasNumberOfSlices = (numberOfSlices: number) => {
+  cy.get('[data-testid="tag-select-slice"] button').click();
```
flash1293, line 102: Could we move the bits relying on specific selectors down into the page to keep the current separation between spec, commands and page?

E.g. 
```
cy.get('[data-testid="tag-select-slice"] button').click();
```

would become `openSlicesDropdown()` implemented in the page

and `cy.get('[data-testid="tag-select-slice"] li')` would become `getSliceDropdownOptions` and so on

airbyte-webapp-e2e-tests/cypress/commands/connectorBuilder.ts
flash1293: You can parse the text as json and deep compare the objects:
```
// getDetectedSchemaElement is implemented in the page, see other comment
getDetectedSchemaElement().should(($el) => {
  expect($el.get(0).innerText).to.eq(SCHEMA_OBJECT)
})
```

airbyte-webapp-e2e-tests/cypress/commands/connectorBuilder.ts
flash1293: Agreed, I think it's OK to just check for "Detected schema and declared schema are different"

flash1293 (commented):
Looks pretty good already, left some nits and moved some testids around.

flash1293:
> I had some flakiness on many type and check methods and ended up adding a lot for { force: true } which feels like a sloppy solution

Good callout, we should look into it. However, I think for the kind of safety we aim to get from these tests it's OK.

airbyte-webapp-e2e-tests/cypress/pages/connectorBuilderPage.ts
maxi297: As the focus is on page 1, there are two elements with "rel=next": page 2 and ">". Therefore, we can't use this solution even though it would have been cleaner

flash1293 (approved):
One small last comment, looks fine to otherwise - no review needed after fixing that

airbyte-webapp-e2e-tests/cypress/pages/connectorBuilderPage.ts
girarda: @maxi297 can you add a comment saying that
> if the arrows placement is changed, this test will fail

girarda (approved):
this is very cool. The PR looks good but I don't feel super qualified to a

octavia-approvington:
This looks fine!  <br/> Merged! <br/> Imagine it being fine

octavia-approvington (approved):

PR #60, opened 2023-02-21
🪟 🧹 Use content hash for build hash

timroes:
## What

As part of the next mono repo step, the OSS folder will move into the `airbyte-platform-internal`. Thus every PR will trigger a rebuild that deploys to cloud, which means regenerating the build hash that indicates whether the FE "please reload the page" notification shows would trigger way too often.

## How

This changes the logic to actually calculate a content hash of all the git tracked files in airbyte-webapp as well as all the orval auto generated files. That hashes will then be converted into one single hash which will be used as a build hash, i.e. touching any orval generated file or airbyte-webapp git commited file will generate a new build hash, but not every rebuild anymore.

## How to test?

To start the server:

```sh
pnpm build
npx http-server build/app
```

Then if you have the webapp open from the `http-server`, running any new `pnpm build` should not show the notification, unless you actually change a file and run `pnpm build`.


airbyte-webapp/scripts/calculate-source-hash.sh
```diff
@@ -0,0 +1,18 @@
+#! /bin/sh
+
+# Script that calculates the hash of all git tracked files as well as the files generated by orval
+
+# Calculate the hash of all files we gitignored in /src/core/request i.e. all the auto generated Orval files.
+GENERATED_CLIENT_HASHES="$(cat .gitignore | grep '/src/core/request' | xargs -I{} openssl sha1 .{})"
```
timroes, line 6: :information_source: Using `openssl sha1` since OSX and Linux have different set of sha tools installed by default, so sha1sum isn't on OSX in the default installation.

josephkmh (approved):
Tested locally and on CI, looks good 👍

octavia-approvington:
This looks fine!  <br/> Merged! <br/> Imagine it being fine

octavia-approvington (approved):

PR #64, opened 2023-02-21
use PAT instead of GITHUB_TOKEN for test_reporter action

cpdeethree:
## What
By default this action uses GITHUB_TOKEN which has a pretty low rate limit. Sometimes it fails for exceeding that rate limit

## How
use env.PAT which is derived from find a pat script instead

source: https://github.com/airbytehq/airbyte/pull/22332


cpdeethree:
Testing a comment 

source: https://github.com/airbytehq/airbyte/pull/22332#issuecomment-1438819464

PR #66, opened 2023-02-21
remove tasks that moves api docs into docs/ (not relevant in new repo)

cgardens:
## What
* In `airbytehq/airbyte` we moved the api docs into the `docs/` folder so that it could be show in docs.airbyte.com. That's not relevant anymore and docs.airbyte.com does not use these docs anymore anyway.

### DIFF
--- a/airbyte-api/build.gradle
+++ b/airbyte-api/build.gradle
@@ -163,19 +163,8 @@ task generateApiDocs(type: GenerateTask) {
             generatePom  : "false",
             interfaceOnly: "true"
     ]
-
-    doLast {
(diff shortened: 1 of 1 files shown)


timroes (approved):
Code LGTM. Have not tested locally, and don't have enough context to judge if those files are no longer needed in docs.

cpdeethree (approved):
if we don't need it in airbytehq/airbyte anymore either, we could probably delete it there too. But that will probably get canned as part of my other deletions in Phase 3 (airbyte-api and octavia are going away over there in phase 3) so I wouldn't worry about it in connectors repo

PR #68, opened 2023-02-21
ensure tasks do not already exist before creating

colesnodgrass:
## What
- inline `publish-repositories.gradle`
    - the `rootDir` path causes an issue when building from cloud
    - saw no benefit to having this maven configuration block existing in a separate file  
- fix airbytehq/airbyte-platform-internal#4474
    - as these tasks will already exist when building built by cloud, check if they already exist before adding them again

## How
- check for task already existing with the same name for tasks `copyDocker` and `listAllDependencies`


colesnodgrass:
This is a [recreated pr](https://github.com/airbytehq/airbyte/pull/23111) that was already approved.

benmoriceau (approved):

PR #79, opened 2023-02-21
add activity to fetch feature flags

pedroslopez:
## What

Migrated from https://github.com/airbytehq/airbyte/pull/22960

In order to be able to use feature flags within the ConnectionManagerWorkflow, we need to introduce a new activity that fetches the feature flags. This is because we can't directly use the feature flag client in the workflow due to the deterministic constraints of temporal workflows.

This will be used to add a feature flag for https://github.com/airbytehq/airbyte/pull/21899

## How

- Add a new FeatureFlagFetchActivity that runs as the first activity in the ConnectionManagerWorkflow
- Set up the activity to fetch some feature flags. Right now this is looking up an unrelated feature flag, but it used to set up the plumbing to be able to fetch an actually useful flag.
- Update tests

### DIFF
--- a/airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/FeatureFlagFetchActivityImpl.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/FeatureFlagFetchActivityImpl.java
@@ -4,8 +4,8 @@
 
 package io.airbyte.workers.temporal.scheduling.activities;
 
-import io.airbyte.api.client.AirbyteApiClient;
 import io.airbyte.api.client.generated.WorkspaceApi;
+import io.airbyte.api.client.invoker.generated.ApiException;
 import io.airbyte.api.client.model.generated.ConnectionIdRequestBody;
 import io.airbyte.api.client.model.generated.WorkspaceRead;
 import io.airbyte.featureflag.FeatureFlagClient;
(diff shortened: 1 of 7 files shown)


pedroslopez:
This PR has been approved on airbytehq/airbyte repo but tagging reviewers to re-approve after migrating this over to airbyte-platform.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/FeatureFlagFetchActivityImpl.java
```diff
@@ -0,0 +1,59 @@
...
+  @Override
+  public FeatureFlagFetchOutput getFeatureFlags(final FeatureFlagFetchInput input) {
+    final UUID workspaceId = getWorkspaceId(input.getConnectionId());
+
+    // TODO: remove this feature flag from here - not really needed by consumers but in here to get this
+    // activity up and running
+    final List<Flag> workspaceFlags = List.of(FieldSelectionEnabled.INSTANCE);
+    final Map<String, Boolean> featureFlags = new HashMap<>();
+    for (final Flag flag : workspaceFlags) {
+      featureFlags.put(flag.getKey(), featureFlagClient.enabled(flag, new Workspace(workspaceId)));
```
benmoriceau, line 53: Is it possible to do a bulk fetch? That would limit the number of calls we will do here. I don't know if there is any limitation with the lauch darkly API but this will be call as many time as we have a sync that runs.

benmoriceau: For later I believe that we should get those through an API, the same way we do for the DB. @colesnodgrass do you think that this would be too complex?

airbyte-workers/src/test/java/io/airbyte/workers/temporal/scheduling/activities/FeatureFlagFetchActivityTest.java
```diff
@@ -0,0 +1,51 @@
...
+  final static UUID WORKSPACE_ID = UUID.randomUUID();
+
+  FeatureFlagFetchActivity featureFlagFetchActivity;
+  FeatureFlagClient featureFlagClient;
+
+  @BeforeEach
+  void setUp() throws ApiException {
+    final WorkspaceApi workspaceApi = mock(WorkspaceApi.class);
+
+    featureFlagClient = new TestClient(Map.of(FieldSelectionEnabled.INSTANCE.getKey(), true));
```
benmoriceau, line 35: NIT, you could use `mock(TestClient.class)`.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/FeatureFlagFetchActivityImpl.java
pedroslopez: My understanding of the feature flag client is that each evaluation of a feature flag does not represent an API call to LaunchDarkly since flags are cached by the SDK https://docs.launchdarkly.com/sdk/concepts/client-side-server-side#server-side-sdks-1

> Server-side SDKs receive the complete ruleset associated with an SDK key when they initialize a connection to LaunchDarkly's servers. LaunchDarkly continuously updates the SDK's cached flag ruleset whenever flag rules change on LaunchDarkly, using this persistent connection.
> 
> When your application requests flag evaluation, server-side SDKs evaluate feature flags using their cached ruleset, and determine the flag variation for a given evaluation context. To do this, they execute an in-process flag evaluation algorithm and return the resulting value.

evantahler (approved):
Approving because https://github.com/airbytehq/airbyte/pull/22960 was already approved

octavia-approvington:
Lets merge it <br/> sheep thumbs up

octavia-approvington (approved):

PR #85, opened 2023-02-21
Extend state aggregator

gosusnp:
## What

Extend StateAggegrator interface:
* Add `isEmpty()` which should tell whether the aggregator contains states
* Add `ingest(StateAggregator other)` to inges all state messages from the other aggregator into the current one.

Relates to #23046 

## How

Due to the recursive nature of how we implemented the interface, extra checks have been added into the `SingleStateAggregator` and `StreamStateAggregator` to fail explicitly in case of unexpected behavior.
The only valid use cases should be when merging data from aggregators of the same nature.


benmoriceau (approved):
LGTM, I wonder why it is needed

gosusnp:
@benmoriceau, that is to prepare for https://github.com/airbytehq/airbyte-platform/pull/86 which is going to be consuming those new methods.

octavia-approvington:
Lets do it! <br/> giddddy-up

octavia-approvington (approved):

PR #86, opened 2023-02-21
Add sync persistence for states

gosusnp:
## What

Add SyncPersistence implementation.
This contains the main logic and the tests, it is not yet plugged.
There will be a subsequent PR to start using this new class. It will come with the related Micronaut changes for injection, the metrics for observability and should be gated by a feature flag.

Closes https://github.com/airbytehq/airbyte/issues/23046
Depends on https://github.com/airbytehq/airbyte-platform/pull/85


airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/sync_persistence/SyncPersistenceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+  private final StateApi stateApi;
+  private final StateAggregatorFactory stateAggregatorFactory;
+
+  private StateAggregator stateBuffer;
+  private StateAggregator stateToFlush;
+  private final ScheduledExecutorService stateFlushExecutorService;
+  private ScheduledFuture<?> stateFlushFuture;
+  private final long stateFlushPeriodInSeconds;
+
+  public SyncPersistenceImpl(final UUID connectionId,
```
gosusnp, line 41: looking into options to not have connectionId

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/sync_persistence/SyncPersistenceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+  public void close() {
+    // stop the buffered refresh
+    stateFlushExecutorService.shutdown();
+
+    // Wait for previous running task to terminate
+    try {
+      final boolean terminated = stateFlushExecutorService.awaitTermination(FLUSH_TERMINATION_TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
+      if (!terminated) {
+        // Ongoing flush failed to terminate within the allocated time
+        log.info("Pending persist operation took too long to complete, most recent states may have been lost");
```
benmoriceau, line 94: Nit: this should be an error log

benmoriceau (approved):
Discussed some points on zoom

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/sync_persistence/SyncPersistenceImpl.java
gosusnp: I thought about it, I think it should be monitored, my argument for not logging an error is that this isn't actionable and it draws attention towards the wrong things. The outcome of a failure here is that we are going to have duplicated data in the raw tables which is going to happen regardless without a bigger change in how we approach state messages.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/sync_persistence/SyncPersistenceImpl.java
gosusnp: Updated

gosusnp:
@benmoriceau , I just updated the PR to avoid passing the `connectionId` on the constructor. It simplifies injection, to avoid misusing the object, I added a check to make sure we do not mix states from different connection.

benmoriceau (approved):

gosusnp:
Superseeded by https://github.com/airbytehq/airbyte-platform-internal/pull/4743

PR #87, opened 2023-02-21
🪟 🎨 🐛 Fix long name overflow and field path names in Catalog Diff modal

timroes:
Migration of https://github.com/airbytehq/airbyte/pull/22610

## What
Related to https://github.com/airbytehq/airbyte/issues/22493

This change mainly fixes a few of the issues in the catalog diff modal

* Long names for namespace, stream name, or field name overflowing to the next column. Now, these fields will be truncated with an ellipsis
* Field names were not showing the full path with dot notation, which made it confusing. For example, instead of showing that "user.name" and "company.name" were added to the stream, it would show the last part of the path so you would see what seems to be a duplicate "name" entry.
* Some of the column names were misaligned, and now they are not

Before:
image

After:
Screen Shot 2023-02-08 at 16 52 26

## How
* Lots of CSS work
* Added mixin for text overflow issues
* Updated storybook to include long names as examples

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.module.scss
+++ b/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.module.scss
@@ -1,5 +1,6 @@
 @forward "./StreamRow.module.scss";
 @forward "./DiffSection.module.scss";
+@use "scss/colors";
 @use "scss/variables";
 
 .row {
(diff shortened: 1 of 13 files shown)


octavia-squidington-iii (approved):
Wow, that for sure looks way better now. Approving since KC and Vlad already approved the original PR.

octavia-approvington (approved):

PR #91, opened 2023-02-21
fix(platform): use updated configuration after check during syncs

pedroslopez:
## What

Migrated from https://github.com/airbytehq/airbyte/pull/21899

fixes https://github.com/airbytehq/airbyte/issues/20912

This PR builds on the changes introduced in https://github.com/airbytehq/airbyte/pull/21629. While the aforementioned PR addressed using updated state/config values _between_ attempts, this PR solves the issue of updating configuration _within an attempt_. This can happen when, for example, an oauth token is refreshed during the `check` operation that can run prior to the SyncWorkflow.

## How

This implements the second part of the solution as laid out in the [tech spec](https://docs.google.com/document/d/1WSZQg2bhNyQD4cgy-Owzejfm8mXhZSYmaqWHlAOqtAs/edit#), changing _when_ we're generating the sync input and _how_ we are generating the check inputs. 

**Before** this change, the SyncWorkflow input was being generated at the start of the ConnectionManagerWorkflow, and was then converted into Check inputs if needed. The activities look something like this:

image

(these are snippets of diagrams, the full views can be found in the tech spec)

**After** this change, we separate the check input generation into its own activity method. Then, we can move the SyncWorkflow input generation to after the check activity runs. Because the previous PR updated the SyncWorkflow input generation activity to fetch latest configs, this will take into consideration any changes that happened during check.

This ends up looking something like this:

image

The change is versioned, so ConnectionManagerWorkflows that executed before this change will still follow the old path/behavior.

## 🚨 User Impact 🚨
No breaking change. Configs updated during check will now correctly be retrieved for the sync.


pedroslopez:
This PR is pending an addition of a feature flag using the changes from #79 , but reviewers, please review away!

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
```diff
@@ -104,105 +108,220 @@ private void saveAttemptSyncConfig(final long jobId, final int attemptNumber, fi
...
+    final IntegrationLauncherConfig sourceLauncherConfig = new IntegrationLauncherConfig()
+        .withJobId(String.valueOf(jobId))
+        .withAttemptId((long) attempt)
+        .withDockerImage(config.getSourceDockerImage())
+        .withProtocolVersion(config.getSourceProtocolVersion())
+        .withIsCustomConnector(config.getIsSourceCustomConnector())
+        .withAllowedHosts(configReplacer.getAllowedHosts(sourceDefinition.getAllowedHosts(), sourceConfiguration));
+
+    if (!ConfigType.RESET_CONNECTION.equals(configType)) {
+      sourceLauncherConfig.setAllowedHosts(configReplacer.getAllowedHosts(sourceDefinition.getAllowedHosts(), sourceConfiguration));
```
gosusnp, line 129: This looks redundant, I think we can remove this if.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
```diff
@@ -104,105 +108,220 @@ private void saveAttemptSyncConfig(final long jobId, final int attemptNumber, fi
         "set attempt sync config");
   }
 
+  private IntegrationLauncherConfig getSourceIntegrationLauncherConfig(final long jobId,
```
gosusnp, line 111: This whole part about building config feels duplicated (from `DefaultJobCreator` maybe?). Is it possible to share this code with the current version to avoid duplication?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
```diff
@@ -208,18 +213,28 @@ private CancellationScope generateSyncWorkflowRunnable(final ConnectionUpdaterIn
...
-        final SyncCheckConnectionFailure syncCheckConnectionFailure = checkConnections(jobInputs);
-        if (syncCheckConnectionFailure.isFailed()) {
-          final StandardSyncOutput checkFailureOutput = syncCheckConnectionFailure.buildFailureOutput();
+        final SyncCheckConnectionResult syncCheckConnectionResult = checkConnections(getJobRunConfig(), jobInputs);
+        if (syncCheckConnectionResult.isFailed()) {
+          final StandardSyncOutput checkFailureOutput = syncCheckConnectionResult.buildFailureOutput();
           workflowState.setFailed(getFailStatus(checkFailureOutput));
           reportFailure(connectionUpdaterInput, checkFailureOutput, FailureCause.CONNECTION);
         } else {
+          if (generateCheckInputVersion >= GENERATE_CHECK_INPUT_CURRENT_VERSION) {
```
gosusnp, line 245: Feels like this try-clause keep growing, I suspect the only part we really want in there is `runChildWorkflow`, probably out of scope for this PR.

gosusnp (commented):
Looks pretty close, I think we have some code duplication in how we build inputs, wondering if we could avoid some of it.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
pedroslopez: Good catch, it wasn't redundant but the mistake is actually the line above. We don't want to set allowed hosts on resets - see https://github.com/airbytehq/airbyte/pull/21935

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
pedroslopez: Hmm, IntegrationLauncherConfig for syncs are only built here - I don't see this in DefaultJobCreator. Could you clarify which duplication you're referring to?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
pedroslopez: I think you're right, but agree it's out of scope for this one

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
```diff
@@ -213,24 +218,32 @@ private CancellationScope generateSyncWorkflowRunnable(final ConnectionUpdaterIn
         prepareForNextRunAndContinueAsNew(connectionUpdaterInput);
       }
 
-      // TODO (pedroslopez): The feature flags will actually be used in a future PR
-      Map<String, Boolean> featureFlags = getFeatureFlags(connectionUpdaterInput.getConnectionId());
+      final Map<String, Boolean> featureFlags = getFeatureFlags(connectionUpdaterInput.getConnectionId());
+      workflowInternalState.setFeatureFlags(featureFlags);
```
pedroslopez, line 222: Note: in order to more easily use feature flags I've added them to the internal workflow state

pedroslopez:
I've now added the feature flag! `connectionManagerWorkflow.checkInputGeneration` will default to false and keep the old ordering of activities / generate check input from the sync input as was done previously, but when enabled will run the new activity for checks and reorder the sync input generation activity to after the checks.

We'll start by rolling this out to some test and later internal workspaces, followed by a full rollout once that's validated after one week. After seeing this live for everyone for another week we'll come back and remove the feature flag.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
pedroslopez: this caused conflicts with replaying old workflows, so did not do this.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
```diff
@@ -218,31 +218,29 @@ private CancellationScope generateSyncWorkflowRunnable(final ConnectionUpdaterIn
         prepareForNextRunAndContinueAsNew(connectionUpdaterInput);
       }
 
-      // TODO (pedroslopez): The feature flags will actually be used in a future PR
-      Map<String, Boolean> featureFlags = getFeatureFlags(connectionUpdaterInput.getConnectionId());
+      final Map<String, Boolean> featureFlags = getFeatureFlags(connectionUpdaterInput.getConnectionId());
+      final boolean featureFlagCheckInputGeneration = featureFlags.getOrDefault(CheckInputGeneration.INSTANCE.getKey(), false);
```
gosusnp, line 222: Feels like this should happen within an activity. My concern of about doing any top level feature flag evaluation is that with temporal replay, it could lead to inconsistencies. Looking up the flag once early and passing the value along should be more predictable.

pedroslopez: > Looking up the flag once early and passing the value along should be more predictable.

This is what we are doing! A new activity was introduced in https://github.com/airbytehq/airbyte-platform/pull/79 to retrieve feature flags for use in the workflow. This happens once at the start.

pedroslopez: getFeatureFlags actually runs the FeatureFlagFetchActivity

https://github.com/airbytehq/airbyte-platform/blob/7f567bb121f44154bb59584d9f9c9381171d265d/airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java#L652-L663

gosusnp: Awesome!

gosusnp (approved):

gosusnp:
LGTM

octavia-approvington:
A crack team of mammals has made a decision.  <br/> imagine a seal of approval

octavia-approvington (approved):

PR #96, opened 2023-02-21
🪟 🎨 Add focus state mixin to Button and Switch components (Migrated from `airbytehq/airbyte`)

tealjulia:
## What
Migrates https://github.com/airbytehq/airbyte/pull/21271

Copied from that PR:

Attempted to address #20150

However, the work has been re-scoped and the original intent of that issue is largely unaddressed here :) 

Instead... we now have a focus ring on our Switch and Button

## Note

During implementation, it was decided that the scope of this work is greater than previously considered and will require reworking how we handle nested interactive elements.  I am leaving the original issue open for that reason.

However, the focus-ring itself is still a useful feature and does not cause problems with the Button and Switch components as implemented.

## How
- adds a mixin that is utilized by Button and can be implemented elsewhere as well
- utilizes the styles directly in Switch due to some implementation differences there

## Recommended reading order
1. `_mixins.scss`
2.  others

*If unsure, leave it blank.*
- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
The user impact is that users will see a focus ring around buttons and switches when said items are focused.


tealjulia:
migrated

PR #98, opened 2023-02-21
Strict comparison normalization: Use launchdarkly feature flag

edgao:
## What
Move https://github.com/airbytehq/airbyte/pull/23045 to the new repo.

flag defined in https://app.launchdarkly.com/default/production/features/normalization.strictComparisonNormalizationEnabled

image tag continues to be an env var though, since it needs to be an arbitrary string

## How
if workspace is in either the env var, _or_ the feature flag, then use the strict_comparison normalization image. This will make the transition smoother (i.e. anyone already in the env var continues to work, and then we can add more workspaces via launchdarkly)

## Recommended reading order
yes, I recommend you read the files in order

## 🚨 User Impact 🚨
none


airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
```diff
@@ -123,7 +129,8 @@ public NormalizationSummary normalize(final JobRunConfig jobRunConfig,
       final var fullDestinationConfig = secretsHydrator.hydrate(input.getDestinationConfiguration());
       final var fullInput = Jsons.clone(input).withDestinationConfiguration(fullDestinationConfig);
 
-      if (FeatureFlagHelper.isStrictComparisonNormalizationEnabledForWorkspace(featureFlags, input.getWorkspaceId())) {
+      if (FeatureFlagHelper.isStrictComparisonNormalizationEnabledForWorkspace(featureFlags, input.getWorkspaceId())
```
grishick, line 132: Just curious - what are the two different scenarios that this `||` condition covers?

edgao: `FeatureFlagHelper` is the legacy-style env var feature flags (hardcoded https://github.com/airbytehq/airbyte-platform-internal/blob/master/infra/gcp/env/prod/helm-data/platform-vars.yaml#L10)

`featureFlagClient` calls LD (using https://app.launchdarkly.com/default/production/features/normalization.strictComparisonNormalizationEnabled/targeting ).

so eventually I want to migrate everything to LD (or at least, any new flags should only use LD). Just didn't want to do that migration in the initial LD rollout.

benmoriceau (approved):

octavia-approvington:
Jerry would be proud <br/> Spongebob

octavia-approvington (approved):

PR #99, opened 2023-02-21
Turn on slack notifications

alovew:
Turn on Slack Notifications for Schema Changes that are detected automatically


gosusnp:
@alovew, what's the impact of this change? are we turning it on for the first time in prod or just changing the default?

alovew:
@gosusnp turning it on for the first time in prod

gosusnp:
> turning it on for the first time in prod

In this case, I think we should update the value set in the charts, not the default. I don't think I review all the pieces related to this change, does this flag only control the addition of a new workflow or will we start notifying users for all changes?
I think we still have some latent schema change issues, https://github.com/airbytehq/oncall/issues/1515 has been added to the backlog, are we going to start notifying those connections?

alovew:
This only controls the workflow - the default is set to true since that's how the other temporal workflow are set. Users can still have the option to turn on or off notifications for their particular connections.

benmoriceau (approved):

gosusnp (approved):

octavia-approvington:
Our work here is done <br/> done

octavia-approvington (approved):

PR #102, opened 2023-02-22
[Migration] [Connector Builder] Separate changes to incorporate Low Code Beta into Connector Builder UI

brianjlai:
## What
Migrated from https://github.com/airbytehq/airbyte/pull/23129 to the `airbyte-platform` repo after the split.

## How
We bumped the `CDK_VERSION` so the UI and builder server use the latest version of the language.

This includes FE changes that add additional transformations of the forms into YAML. There were also some changes to the forms themselves for concepts like stream slicer that were deprecated. The connector builder backend was modified to support the breaking changes to the language.

## Recommended reading order
1. `x.java`
2. `y.java`

- [x] YES 💚
- [ ] NO ❌


airbyte-webapp/src/components/connectorBuilder/Builder/RequestOptionFields.tsx
```diff
@@ -0,0 +1,40 @@
+import { useField } from "formik";
```
brianjlai, line 1: noting that this is the only file that deviated from the original PR. This is because we renamed this file and generating patch files because it can only detect changes in present files and not ones that have been removed or renamed. (something something change data capture analogy)

brianjlai: actually git is smart enough that once i found the extra file and deleted it, it marked this as moved. so we should be good here

brianjlai:
verified on a local Airbyte instance that an incremental sync in both YAML and UI work when testing records

flash1293 (approved):
Tested locally and works as expected

PR #110, opened 2023-02-22
Issue 21933 make connector builder server build incremental

maxi297:
## What
Migration of https://github.com/airbytehq/airbyte/pull/22548


sherifnada (approved):
I think you should close the original issue and mention while we didn't make this completely incremental the build now takes like 4s on the second build and anyways it will become a non issue once we move to micronaut which is happening soon

maxi297:
Has been moved manually like /create-oss-pr would have done

PR #112, opened 2023-02-22
🪟 🎉 Adds a TextWithOverflowTooltip component

tealjulia:
## What
**TL;DR**: Uses `react-resize-detector` and `react-intersection-observer` to decide whether to wrap a Text component in a Tooltip.

## How
- Check if the HTMLElement inside the Text component is in view (using `react-intersection-observer`)
- If it is, compare the scroll width to the client width to determine if the text is overflowed
- If it is, wrap the element in a tooltip
- Re-checks these things if the element is resized (using `react-resize-detector`)

## Recommended reading order
1. `TextWithOverflowTooltip.*`
2. `CatalogTreeTableCell.tsx`
3. `CatalogTreeTableRow.tsx`
4. others

*If unsure, leave it blank.*
- [X] YES 💚
- [ ] NO ❌


airbyte-webapp/src/components/ui/Text/Text.tsx
```diff
@@ -42,29 +43,36 @@ const getTextClassNames = ({
...
-    color = "default",
-    gradient = false,
-    ...remainingProps
-  }) => {
-    const className = classNames(
-      getTextClassNames({ centered, size, bold, inverseColor, gradient, color }),
-      classNameProp
-    );
+export const Text: React.FC<React.PropsWithRef<React.PropsWithChildren<TextProps>>> = React.memo(
+  React.forwardRef(
```
tealjulia, line 47: Ugly diff -- this was wrapped in a forwardRef and the ref was passed down to the text

airbyte-webapp/src/components/ui/PillSelect/PillButton.tsx
```diff
@@ -47,9 +49,15 @@ export const PillButton: React.FC<PillButtonProps> = ({
...
-            </Text>
+            {withOverflowTooltip ? (
+              <TextWithOverflowTooltip as="span" size="xs" className={styles.text}>
+                {child}
+              </TextWithOverflowTooltip>
+            ) : (
+              <Text as="span" size="xs" className={styles.text}>
+                {child}
+              </Text>
+            )}
```
krishnaglick, line 60: I think this is may be the wrong level for the tooltip.
My PK selection cell won't show overflow unless I select 2 or more items.

airbyte-webapp/src/components/ui/Text/TextWithOverflowTooltip.tsx
```diff
@@ -0,0 +1,58 @@
...
+              inViewRef(element);
+            }}
+            className={classNames(className, styles.overflowText)}
+            {...props}
+          >
+            {children}
+          </Text>
+        }
+        disabled={tooltipDisabled}
+        className={styles.fullWidthTooltip} // todo: is that right?
```
krishnaglick, line 51: I'm unsure what logic you're porting from but this will allow the tooltip to be any width.

tealjulia: I pulled this in from somewhere in the catalog tree table row.

Looking at this with fresh eyes... We probably _don't_ want this by default so the tooltip will wrap.

krishnaglick: I'd agree with that. I have a tickling in the back of my head that I applied some width-adjusting style in my table work. Lemme know if you'd like to pair to sort through this!

tealjulia: It was for when things have very long strings with no word break ie:


added a `break-word` instead, which seems cleaner?:

tealjulia:
migrated

airbyte-webapp/src/components/ui/PillSelect/PillButton.tsx
tealjulia: D'oh!  It's the `as="span"`.  Can't overflow text in a span :)

PR #114, opened 2023-02-22
Stream-Centric UI Feature Flag & Status Page Rename

krishnaglick:
## What
Resolves https://github.com/airbytehq/airbyte/issues/23354
Resolves https://github.com/airbytehq/airbyte/issues/23343
Adds a feature flag for stream-centric UI work and sets up the connection detail page for further work.

## How
`connection.streamCentricUI` feature flag added in the code and in [LaunchDarkly](https://app.launchdarkly.com/default/production/features/connection.streamCentricUI/targeting)
Connection Status page renamed to Connection Job History in prep for new Stream-Centric Status Page.


airbyte-webapp/src/pages/connections/ConnectionJobHistory/ConnectionJobHistory.tsx
```diff
@@ -43,7 +43,7 @@ const getJobRunningOrPending = (jobs: JobWithAttemptsRead[]) => {
   });
 };
 
-export const ConnectionStatusPage: React.FC = () => {
+export const ConnectionJobHistory: React.FC = () => {
```
edmundito, line 46: It should be `ConnectionJobHistoryPage/ConnectionJobHistoryPage`. All pages components end with `Page`.

airbyte-webapp/src/pages/connections/ConnectionsRoutes.tsx
```diff
@@ -3,31 +3,40 @@ import { Navigate, Route, Routes } from "react-router-dom";
...
+  const streamCentricUIEnabled = useExperiment("connection.streamCentricUI", false);
+  return (
+    <Suspense fallback={<LoadingPage />}>
+      <Routes>
+        <Route path={RoutePaths.ConnectionNew} element={<CreateConnectionPage />} />
+        <Route path={ConnectionRoutePaths.Root} element={<ConnectionPage />}>
+          <Route
+            path={ConnectionRoutePaths.Status}
+            element={streamCentricUIEnabled ? null : <ConnectionJobHistory />}
+          />
```
edmundito, line 29: Just thinking about this from another way, may be easier to track which routes have changed with the feature:
```tsx
{streamCentricUIEnabled ?
<>
   <Route path={ConnectionRoutePaths.Status} element={null} />
   <Route path={ConnectionRoutePaths.JobHistory} element={<ConnectionHistory />} />
</>
:
   <Route path={ConnectionRoutePaths.Status} element={<ConnectionHistory />} />
}
```

airbyte-webapp/src/pages/connections/types.ts
```diff
@@ -4,4 +4,5 @@ export const enum ConnectionRoutePaths {
   Transformation = "transformation",
   Replication = "replication",
   Settings = "settings",
+  JobHistory = "job-history",
```
edmundito, line 7: Just want to make sure that this name makes sense. The terminology is currently "syncs", and I wonder if this has been thought through roughly before changing the term to jobs.

krishnaglick: I think that's an @andyjih question

andyjih: Job History is correct!

edmundito (approved):

PR #117, opened 2023-02-22
Migration/bmoric/fetch timeout actor configuration

benmoriceau:
## What
* Update the SourceDefinition accessor in order to add the new column seconds_between_messages to the result. This is then propagate to the API which will be use in the worker.
* The default value is 10800 (60 * 60 * 3 = 3 hours in seconds) which is the default value for OSS
* The value is not added to the config on purpose. It is needed on `ServerApp.java` which will be remove soon.

## Recommended reading order
1. `ConfigPersistence.java` => fetch the new column
2. `DatabaseBeanFactory.java` => Inject the default value

From: https://github.com/airbytehq/airbyte/pull/22957


airbyte-bootloader/src/main/resources/application.yml
```diff
@@ -36,6 +36,9 @@ airbyte:
         address: ${VAULT_ADDRESS:}
         prefix: ${VAULT_PREFIX:}
         token: ${VAULT_AUTH_TOKEN:}
+  worker:
```
gosusnp, line 39: May have missed this in the original PR, why do we have worker config in bootloader?

airbyte-cron/src/main/resources/application.yml
```diff
@@ -29,6 +29,8 @@ airbyte:
       ports: ${TEMPORAL_WORKER_PORTS:}
   tracking-strategy: ${TRACKING_STRATEGY:LOGGING}
   version: ${AIRBYTE_VERSION}
+  worker:
```
gosusnp, line 32: Similar question, why do we have worker config in the cron

gosusnp (approved):
LGTM.
Just a minor question, why do we replicate what seems to be worker config in other apps.yml?

PR #118, opened 2023-02-22
[Migration] [Connector Builder] Return string format to allow for connector read() that return non-object responses

brianjlai:
Migrates https://github.com/airbytehq/airbyte/issues/20764 after the platform monorepo split

joe wrote the FE and I wrote the connector builder changes

## What
For APIs with responses that were non-object responses encapsulated in `{ ... }`. The connector builder read() would fail. It is perfectly reasonable for a API to response with an array of elements `[ ... ]` and potentially even string input. This change should allow for both.

## How
The connector builder server no longer parsed response output into a json object because it might not be one. Instead we pass along the string output and let the frontend handle deserializing it back into the appropriate object to display back in the test output window

## Recommended reading order
1. `default_api.py`
2. `PageDisplay.tsx`

- [x] YES 💚

### DIFF
--- a/airbyte-connector-builder-server/connector_builder/generated/models/http_response.py
+++ b/airbyte-connector-builder-server/connector_builder/generated/models/http_response.py
@@ -22,7 +22,7 @@ class HttpResponse(BaseModel):
     """
 
     status: int
-    body: Optional[Dict[str, Any]] = None
+    body: Optional[str] = None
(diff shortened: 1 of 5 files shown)


lmossman (approved):
Tested locally and it works as expected! LGTM

girarda (approved):
🚢

PR #119, opened 2023-02-23
🪟 🎉  List connector builder projects UI

lmossman:
## What
Based off of the `feature/connector-builder-projects-ui` feature branch, which will be a long-lived branch to collect the connector builder project UI changes so that they can be merged in together. That feature branch itself is based off of `flash1293/connector-builder-api-part-2-move` so that the builder projects API endpoints are accessible.

Resolves https://github.com/airbytehq/airbyte/issues/22448

## How
Uses the NextTable component to implement the builder projects listing page in a new `ConnectorBuilderListPage` component.

`ConnectorBuilderLandingPage` was renamed to `ConnectorBuilderCreatePage` and was placed under the `/create` path, with the relevant navigation logic updated.

Screenshot of the new page:
Screen Shot 2023-02-22 at 6 30 19 PM

## Recommended reading order
1. `ConnectorBuilderListPage`
2. `ConnectorBuilderProjectsService`
3. The rest


airbyte-webapp/src/components/ui/NextTable/NextTable.module.scss
```diff
@@ -46,7 +46,7 @@
   }
 
   &:first-child {
-    padding-left: calc(variables.$spacing-2xl + variables.$spacing-sm);
+    padding-left: calc(variables.$spacing-xl + variables.$spacing-sm);
```
lmossman, line 49: I made these changes in this file because the padding on the first child seemed weirdly high to me, and the custom padding in the cell data didn't seem necessary and just resulted in misaligned columns. Here are some screenshots showing what this changes in various tables in the app:

### Sources page
before:
Screen Shot 2023-02-22 at 6 34 03 PM

after:
Screen Shot 2023-02-22 at 6 34 12 PM

### Connections page
before:
Screen Shot 2023-02-22 at 6 35 09 PM

after:
Screen Shot 2023-02-22 at 6 35 28 PM

### Settings > Sources page
before:
Screen Shot 2023-02-22 at 6 36 24 PM

after:
Screen Shot 2023-02-22 at 6 36 42 PM

I think the weird padding rules in these table styles can still be cleaned up more, but I didn't want to spend too much time on that for this PR specifically. Lmk what you think

flash1293: It looks like the first column has a left padding of 32px, set here: https://github.com/airbytehq/airbyte-platform/blob/c6d8f28eafafeceaca7016f3ba0d6caa120172ab/airbyte-webapp/src/components/EntityTable/ImplementationTable.module.scss#L3

However setting the padding left of this to 32px as well doesn't work, because the header text is wrapped in a button and that has a default of 6px by the browser stylesheet:
Screenshot 2023-02-23 at 11 54 49


However, even removing this doesn't fix it everywhere because the connection table overwrites the 32px left padding... No easy win here. FYI @edmundito @josephkmh we should align these padding values for all tables for a consistent look.

airbyte-webapp/src/core/domain/connectorBuilder/ConnectorBuilderProjectsRequestService.ts
```diff
@@ -0,0 +1,8 @@
+import { listConnectorBuilderProjects } from "core/request/AirbyteClient";
+import { AirbyteRequestService } from "core/request/AirbyteRequestService";
+
+export class ConnectorBuilderProjectsRequestService extends AirbyteRequestService {
+  public list(workspaceId: string) {
```
flash1293, line 5: FYI @timroes - you mentioned you wanted to refactor how the api service works, we are currently introducing new functionality here

airbyte-webapp/src/pages/connectorBuilder/ConnectorBuilderListPage/ConnectorBuilderListPage.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export const ConnectorBuilderListPage: React.FC = () => {
+  const navigate = useNavigate();
+  const query = useQuery<{ sortBy?: string; order?: SortOrderEnum }>();
+  const workspaceId = useCurrentWorkspaceId();
+  const projectsReadList = useListProjects(workspaceId);
+  // TODO: set version based on activeDeclarativeManifestVersion once it is added to the API
+  const projects = projectsReadList.projects.map((projectDetails) => ({ name: projectDetails.name, version: "draft" }));
+
```
flash1293, line 39: I wrote this down in a pretty convoluted way in the issue, but if there are no builder projects, it should redirect immediately to `connector-builder/create` (with `replace: true`) - right now it shows an empty table:
Screenshot 2023-02-23 at 12 06 57

flash1293: The e2e tests are currently failing, but I think fixing this issue would fix the tests as well

airbyte-webapp/src/components/ui/NextTable/NextTable.module.scss
josephkmh: Yeah these tables are really due for an overhaul. IIRC `EntityTable` is set to be replaced by `NextTable`, though I'm not sure on the timeline there, and whether any team has replacing these specific tables on their roadmap. @edmundito or @dizel852 do you have more insight here?

lmossman: Okay, sounds like I should go ahead and undo the `NextTable.module.scss` changes I made in this PR and instead just change it for the specific table I'm adding here

airbyte-webapp/src/pages/connectorBuilder/ConnectorBuilderListPage/ConnectorBuilderListPage.tsx
lmossman: Ah yeah I forgot to add this, will do

lmossman:
@flash1293 I've addressed your comments here so this should be ready for another look

flash1293 (approved):
LGTM, thanks for these fixes

PR #123, opened 2023-02-23
🪟 🔧 Replace husky by git core.hooksPath

timroes:
## What

This removes husky and replaces it by setting `git config core.hooksPath` directly to the checked in hooks in `airbyte-webapp/.gitconfig`. 

This also ensures that the new files are working in airbyte-platform (where the path will be `airbyte-webapp/.githooks`) as well as `airbyte-platform-internal` (where the path will be `oss/airbyte-webapp/.githooks`), by making sure it will when installing and when running properly go up or down the repository.

## Testing

To test this in `airbyte-platform-internal`, you can go to the root of the repository there, switch to the `oss-in-cloud` branch and execute the following command to apply the patch of this PR to the repository.

```
curl https://patch-diff.githubusercontent.com/raw/airbytehq/airbyte-platform/pull/123.patch | git apply --directory=oss
```

I've validated that this works in both repositories, as well as via `pnpm` or `gradle` when running. Happy to get revalidation on that.


josephkmh (approved):
Looks good, tested in both `airbyte-internal` and `airbyte-internal-platform` and the precommit hook location was resolved properly in both

timroes:
Merging off cycle, since this is only touching development setup and no production code.

PR #124, opened 2023-02-23
🪟 🎨 Show warning in UI when in workspace that you are not a member of

josephkmh:
Migrated PR: https://github.com/airbytehq/airbyte/pull/23141

## What
Closes https://github.com/airbytehq/airbyte/issues/7614

Displays a prominent red warning when you are in a cloud workspace that you are not a member of. This is useful when Airbyte employees are in customer workspaces.

https://user-images.githubusercontent.com/7550957/220940517-e980629a-5463-46c9-a152-b039dda6c85a.mov

This depends on the FeatureItem.ShowCustomerWorkspaceWarning feature flag, which is currently off by default. After validating this is desirable to display for all Airbyte staff, we could consider removing the flag.

## How
- Adds a new feature flag: `FeatureItem.ShowAdminWarningInWorkspace` (disabled by default)
- When enabled, `AdminWorkspaceWarning` queries members of the current workspace via `/cloud/v1/web_backend/permissions/list_workspaces_by_user` and checks if the current user is one of them

## Recommended reading order
Top to bottom

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌


tealjulia (approved):
Looks fine to me!  Tested locally and it's there when it should be/not when it's not (when I am a creator or invited member of a workspace).  

I wonder if both the popout and this could use the `PillButton` component to avoid hardcoding styles, but that seems like a separate issue! (Maybe `WorkspacesPopout` could utilize `PillSelect` even, though probably a non-trivial change 🤔)

PR #125, opened 2023-02-23
[DRAFT]🪟🚦Add E2E tests for syncModes in new table

edmundito:
## What
Resolves https://github.com/airbytehq/airbyte/issues/22641

Adds E2E tests to verify syncModes in the new streams table table.

## How
TODO


airbyte-webapp-e2e-tests/cypress/integration/connection/syncModes.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+      });
+    });
+
+    describe("with selectable primary keys", () => {
+      const cursorValue = "created_at";
+      const primaryKeyValue = ["car_id", "user_id"];
+
+      before(() => {
+        streamsTable.searchStream("user_cars");
+        streamsTable.selectSyncMode("Incremental", "Deduped + history");
```
dizel852, line 281: 💡 Just an idea: would be great to access sync modes via key. We can convert it to a map with tuples, for example:
const syncModes = {
      INCR_DEDUP = ["Incremental", "Deduped + history"]
      .....
}
```suggestion
        streamsTable.selectSyncMode(syncModes.INCR_DEDUP);
```

airbyte-webapp-e2e-tests/cypress/integration/connection/syncModes.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+      });
+
+      it("has empty cursor and primary key selects", () => {
+        streamsTable.hasEmptyCursorSelect("public", "user_cars");
+        streamsTable.hasEmptyPrimaryKeySelect("public", "user_cars");
```
dizel852, line 286: 💬 Discussion question:
Also faced the same flow - we always need to provide a namespace and stream name to have the stream we are looking for. And since in most of the cases we are working just with one stream, I decided to try to solve it via closure, for example:
```
const {toggleStreamSync, isStreamSyncEnabled, isStreamRowViisible} = getStreamUtitilyFunctions("namespace", "streamname")
```

Raw example - then, in tests we just need:

```
it("...",() =>{
    toggleStreamSync()
    isStreamSyncEnabled(true)
} )

it("...",() =>{
    toggleStreamSync()
    isStreamSyncEnabled(false)
} )
```

WDYT?

dizel852 (approved):
LGTM 👍 Also tested locally

Left some comments, they are more like a discussion.

krishnaglick (approved):
Had one minor comment.

edmundito:
Rebased the changes with Cypress 12 and re-formatted the test to run in isolation which is now the default.

https://docs.cypress.io/guides/core-concepts/test-isolation

airbyte-webapp-e2e-tests/cypress/integration/connection/syncModes.spec.ts
edmundito: I had some ideas on how to improve this but decided to stick with what we have available because the changes would also impact `connection.cy.ts`. I agree that it's better to just work within the context of the stream instead of calling the same thing repeatedly but that's something that I'd like to refactor later.

airbyte-webapp-e2e-tests/cypress/integration/connection/syncModes.spec.ts
edmundito: Good idea. I will push a change

edmundito:
Final changes addressing @dizel852 's suggestion to treat the sync mode names through constant values. I opted to use enums.

airbyte-webapp-e2e-tests/cypress/integration/connection/syncModes.spec.ts
dizel852: Oh, okay, gotcha 👌

dizel852 (approved):
Awesome 👍

PR #126, opened 2023-02-23
Migration/bmoric/fail connection if heartbeat failure

benmoriceau:
## What
This is checking than the source is pushing data or not. If the source haven't emitted any message within the configured acceptable delay, the connection will be failed.

The beat of the heartbeat monitor is done in the `DefaultAirbyteSource.java`. This allow to include the logs message as a message relevant for the timeout. Then another class `HeartbeatTimeoutChaperon` is use to monitor that the timeout is up-to-date. It will check every minute if the source is still beating which means that it has send a message in the last X seconds (X is a configurable value).
The heartbeat timeout is a wrapper of the execution. It will start 2 threads, one to run the timeout check, one to run the runnable provided to it. It will then wait until one of the thread terminate. If the timeout check thread is the first to finish, it means that the source is not responsive which will make the connection to fail with a dedicated failure type.

There is a feature flag which prevent making a connection to fail, if the flag is false, the chaperone will log that the timeout check failed won't return so the sync won't be failed. There is another PR that will add datadog monitoring when that happens.

## Recommended reading order
1. `HeartbeatTimeoutChaperone.java` => wrapper to check the heartbeat
2. `DefaultReplicationWorker.java` => Call the chaperone and create a custom error messages
3. `ReplicationJobOrchestrator.java` and `ReplicationActivityImpl.java` => Load the DB configured timeout value through the api
4. `FailureHelper.java` => Error message that will be displayed to the user.

Source: https://github.com/airbytehq/airbyte/pull/22397


airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -215,8 +220,8 @@ private void replicate(final Path jobRoot,
             }
           });
 
-      final CompletableFuture<?> readSrcAndWriteDstThread = CompletableFuture.runAsync(
-          readFromSrcAndWriteToDstRunnable(
+      final CompletableFuture<?> readSrcAndWriteDstThread = srcHeartbeatTimeoutChaperone
```
benmoriceau, line 223: @gosusnp I was needing a CompletableFuture here because of the log on line 250 and 252.

benmoriceau:
@gosusnp I have address what we discussed on zoom

gosusnp (approved):
LGTM

erohmensing:
@benmoriceau looks like merging master in caused some diff issues 😅

benmoriceau:
> @benmoriceau looks like merging master in caused some diff issues 😅

I will check with Conor

erohmensing:
Removed connector ops' request as it seems it was added automatically before, and there aren't any changes to Oauth. Feel free to re-request if you actually want one from us!

PR #127, opened 2023-02-23
Add datadog for postgres source

sashaNeshcheret:
## What
*Describe what the change is solving*
*It helps to add screenshots if it affects the frontend.*

## How
*Describe the solution*

## Recommended reading order
1. `x.java`
2. `y.java`

*If unsure, leave it blank.*
- [ ] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨


airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
```diff
@@ -41,6 +41,10 @@ public class DockerProcessFactory implements ProcessFactory {
   private static final Path DATA_MOUNT_DESTINATION = Path.of("/data");
   private static final Path LOCAL_MOUNT_DESTINATION = Path.of("/local");
   private static final String IMAGE_EXISTS_SCRIPT = "image_exists.sh";
+  private static final String DD_ENV_VAR =
+      "-XX:+ExitOnOutOfMemoryError -javaagent:/airbyte/dd-java-agent.jar -Ddd.profiling.enabled=true -XX:FlightRecorderOptions=stackdepth=256 -Ddd.trace.sample.rate=30 -Ddd.trace.request_header.tags=User-Agent:http.useragent";
+  private static final List<String> DATADOG_SUPPORT_IMAGES = List.of("source-postgres");
```
jdpgrailsdev, line 46: For future consideration:  we should probably move this list out into an env var to make it easy to update if/when we add more connectors to the ones that we want to enable metrics for.

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
```diff
@@ -116,6 +116,14 @@ public class KubePodProcess implements KubePod {
   private static final ResourceRequirements DEFAULT_SOCAT_RESOURCES = new ResourceRequirements()
       .withMemoryLimit(configs.getSidecarKubeMemoryLimit()).withMemoryRequest(configs.getSidecarMemoryRequest())
       .withCpuLimit(configs.getSocatSidecarKubeCpuLimit()).withCpuRequest(configs.getSocatSidecarKubeCpuRequest());
+  private static final String DD_ENV_VAR =
+      "-XX:+ExitOnOutOfMemoryError -javaagent:/airbyte/dd-java-agent.jar -Ddd.profiling.enabled=true -XX:FlightRecorderOptions=stackdepth=256 -Ddd.trace.sample.rate=30 -Ddd.trace.request_header.tags=User-Agent:http.useragent";
```
jdpgrailsdev, line 120: This is duplicated between `DockerProcessFactory` and `KubeProdProcess`.  Consider extracting to a common place so we don't have to remember to update it in two places should the command change.

jdpgrailsdev (commented):
LGTM.  There are a couple of small items to address and it looks like there is a checkstyle failure with the PR build.

PR #128, opened 2023-02-23
Update to latest jOOQ release

jdpgrailsdev:
## What
* Update to the latest version of jOOQ to potentially address memory leaks

## How
* Bump jOOQ from 3.13.4 -> 3.17.8
* Bump platform Postgresql test container to 1.17.6 to match version provided by Micronaut
* Force Postgresql test container version to ensure consistency with schema dump
* Address jOOQ related changes (`DSLContext` is no longer auto-closeable, other API changes)

## Recommended reading order
1. `deps.toml`
2. `airbyte-db/db-lib/build.gradle`
3. `airbyte-db/jooq/build.gradle`
4. `DefaultJobPersistence.java`

- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
None

## Tests

All existing tests pass without modification to the assertions.

**N.B.**:  Plan is to test a build with this PR to see if it actually addresses the memory leak before merging.


airbyte-db/db-lib/src/main/resources/configs_database/schema_dump.txt
```diff
@@ -1,209 +1,205 @@
...
   "actor_definition_id" uuid not null,
   "name" varchar(256) not null,
   "configuration" jsonb not null,
-  "actor_type" actor_type not null,
-  "tombstone" bool not null default false,
-  "created_at" timestamptz(35) not null default null,
-  "updated_at" timestamptz(35) not null default null,
+  "actor_type" "public"."actor_type" not null,
+  "tombstone" boolean not null default false,
+  "created_at" timestamp(6) with time zone not null default current_timestamp,
```
gosusnp, line 13: Is this type change expected? It feels more than just cosmetic.

airbyte-db/db-lib/src/main/resources/configs_database/schema_dump.txt
```diff
@@ -1,209 +1,205 @@
...
+  "send_newsletter" boolean,
+  "send_security_updates" boolean,
+  "display_setup_wizard" boolean,
+  "tombstone" boolean not null default false,
+  "notifications" jsonb,
+  "first_sync_complete" boolean,
+  "feedback_complete" boolean,
+  "created_at" timestamp(6) with time zone not null default current_timestamp,
+  "updated_at" timestamp(6) with time zone not null default current_timestamp,
+  "geography" geography_type not null default cast('AUTO' as geography_type),
```
gosusnp, line 188: Same question, sanity checking here.

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/DefaultJobPersistence.java
```diff
@@ -1378,8 +1386,10 @@ private static Object getJsonNodeValue(final JsonNode jsonNode, final String col
       return valueNode.asDouble();
     } else if (nodeType == JsonNodeType.NULL) {
       return null;
+    } else if (nodeType == JsonNodeType.MISSING) {
```
gosusnp, line 1389: Did you run into valid cases where we'd hit nodeType is MISSING? Not sure when that would happen, I am wondering if throwing an error could be safer.

gosusnp (approved):
Looks good, just noted a few places where it could be worth double checking.

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/DefaultJobPersistence.java
```diff
@@ -1212,19 +1213,26 @@ public void purgeJobHistory(final LocalDateTime asOfDate) {
...
+      final Set<String> jsonFieldNames = Arrays.stream(record.fields())
+          .filter(f -> "jsonb".equals(f.getDataType().getTypeName()))
+          .map(Field::getName)
+          .collect(Collectors.toSet());
+      final String json = record.formatJSON(JdbcUtils.getDefaultJSONFormat());
+      final JsonNode row = Jsons.deserialize(json);
+      // for json fields, deserialize them so they are treated as objects instead of strings. this is to
+      // get around that formatJson doesn't handle deserializing them for us.
+      jsonFieldNames.forEach(jsonFieldName -> {
+        // Ensure that missing fields are converted into an empty JSON object in order to pass JASON
```
colesnodgrass, line 1227: typo in comment - `JASON`

colesnodgrass (commented):
Code changes seem fine.  I have the same concerns as Jimmy about the types changing on the db-dump.

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/DefaultJobPersistence.java
jdpgrailsdev: Thanks...will fix

airbyte-db/db-lib/src/main/resources/configs_database/schema_dump.txt
jdpgrailsdev: Unsure.  We are two years behind in terms of jOOQ releases, so its not surprising that they changed the datatypes oer that period.

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/DefaultJobPersistence.java
jdpgrailsdev: The `DefaultJobPersistenceTest#Should have valid yaml schemas in exported database` test failed because the `JsonNodeType.MISSING` was not in that if/else statement, resulting in it falling through to the exception below it.  The test passed before the update, so I added this to maintain the test.  Whether or not the test is valid is a separate discussion.

gosusnp: This is testing `jobPersistence.exportDatabase()` which I believe has been deprecated. @davinchia, may be you can help confirm this.
It might be better to delete the whole test rather than be more permissive regarding unexpected Json data.

airbyte-db/db-lib/src/main/resources/configs_database/schema_dump.txt
benmoriceau: could this have an impact on the new DB? https://medium.com/building-the-system/how-to-store-dates-and-times-in-postgresql-269bda8d6403#:~:text=timestamp%20vs.-,timestamptz,%E2%80%9Ctimestamp%E2%80%9D%20doesn't.&text=From%20the%20previous%20example%3A,')%20from%20the%20original%20value. This is what I found to explain de difference between timestampz and timestamp. It feel that we are changing the type of the column and I'm not sure about what could be the consequences on this

jdpgrailsdev: To be clear, this is just changing in the those dump files, which are made from a new containerized database migrated from scratch.  The question is whether or not the generated jOOQ classes will be compatible with the existing database.  We won't know for sure until we test this in Cloud in one of the dev environments.

jdpgrailsdev: @gosusnp @benmoriceau I looked into this some more and I don't believe that the timestampz -> timestamp with time zone is an issue.  The generated jOOQ model code still uses the same datatype.  This really just impacts how jOOQ interprets the schema when exporting it to a file.  The type in the database is still the same as before, as it utilizes Flyway to actually create/migrate the database.  I suppose the issue here is only for those who attempt to re-create the database from the dumped schema file instead of using the Bootloader to manage the creation and migration of the database.  Therefore, I don't think that there is much risk here.

gosusnp: Thanks, sounds like as long as we make sure to have a test with a rolling upgrade, we should be good then.

jdpgrailsdev:
Closing in favor of internal PR

PR #129, opened 2023-02-23
🪟 🐛 [Connector Builder] fix label rendering issue when switching between oneOf types

lmossman:
## What
An issue I noticed while investigating a separate problem was that the label of the `paginator.strategy.page_size` field was not being updated when switching between Offset and Page Increment pagination strategies in the builder.

Here is a quick screen recording demonstrating the weird behavior. Notice that if I switch to Cursor pagination, then back to one of the above two, it fixes the label on the first type switch, but then keeps that same label when switching between the other two types.

https://user-images.githubusercontent.com/22731524/221055861-9bed3d72-76ff-4e14-85be-5fc3422c5bbc.mov

## How
This issue seems to be a regression introduced by the switch to FastFields. The problem is that the BuilderField's FastField component is currently only being re-rendered when either the `path` or the `type` of the field are modified. In this case, the path is always staying the same (it is always `paginator.strategy.page_size` regardless of the pagination strategy), and the type is always `number`.

The fix here is to add the field's `label` to the FastField key, so that it is re-rendered if the label changes too.


flash1293 (approved):
Tested and works as expected, LGTM. Thanks for fixing this!

PR #130, opened 2023-02-24
Default to empty list for airgapped Oss Store reference

bnchrch:
## What
In this classes previous life as [AirbyteGithubStore](https://github.com/airbytehq/airbyte/blob/v0.40.32/airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteGithubStore.java)

We caught any exception and returned an empty list.

This allowed us to support when perhaps the server didnt have internet access (airgapped for example)

## How
This re-adds this catch at the method level

### Fallback to Database did not exist
@pedroslopez the previous error message
> Unable to retrieve latest Destination list from Github. Using the list bundled with Airbyte. This warning is expected if this Airbyte cluster does not have internet access.

Indicates that we fell back to the list bundled with airbyte. Looking at the code that was not the case. Perhaps was at some point it doesnt seems so now.

Am I wrong there? If not should we add this functionality?

Would be out of scope for this fix but I do feel like a local provider fallback in remote could be useful.

### Could not reproduce
@flash1293 I wasn't able to reproduce the error [you had mentioned here](https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1677167141062969) 

Looking at the Error it seems like micronaut may have reset the connection?
> "Caused by: java.io.IOException: Connection reset",

Unsure what couldve done that.


flash1293:
I'm not sure the empty list is helpful in this case - the UI would be rendered pretty useless in this case as no sources/definitions could be configured. The fallback to a local definition list seems much better.

Another thing - it seems like on every refresh of the webapp (which triggers a request to `http://127.0.0.1:8001/api/v1/source_definitions/list_for_workspace`) it seems like the server is _refetching_ the list from the remote URL. This could result in a ton of requests to the definitions file, is this working as expected?

pedroslopez (approved):
I think returning an empty list here is fine and matches the previous behavior. In practice all this means is when going to the settings page we won't see the what the latest version is for these sources, which makes sense if you don't have internet access. 

The list of sources/destinations available is still populated by what's in the DB, the frontend just fills in the latest version for each using the information returned by this endpoint. 

See how frontend handles it here:
https://github.com/airbytehq/airbyte-platform/blob/f9c57ffa91fe9446843f54326f58d44ca349c246/airbyte-webapp/src/services/connector/SourceDefinitionService.ts#L45-L54

The "refresh on each request" behavior also matches AirbyteGithubStore, but agree that we may be able to cache things for a bit, specially reusing the `@Cacheable` behavior from micronaut, but out of scope for now.

Note that this change in behavior might mean that we _can't_ use RemoteDefinitionsProvider directly as was planned before.

flash1293:
Thanks for the context @pedroslopez , that makes a lot of sense

flash1293 (approved):

bnchrch:
Closed in favor of https://github.com/airbytehq/airbyte-platform-internal/pull/4709

PR #132, opened 2023-02-24
Add local catalog path to .env

bnchrch:
## What

We have a way to override the path to the local catalog in oss development.

However there are two issues
1. its not currently documented and is not visible to the average user.
2. It didn't have support for docker-compose

## How
1. This adds the .env.dev reference
2. And also adds the env var to docker-compose.yml


airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
```diff
@@ -19,21 +19,25 @@
...
+
   private static final String LOCAL_CONNECTOR_CATALOG_PATH = CatalogDefinitionsConfig.getLocalConnectorCatalogPath();
 
   public CombinedConnectorCatalog getLocalDefinitionCatalog() {
     try {
       final URL url = Resources.getResource(LOCAL_CONNECTOR_CATALOG_PATH);
-      final String jsonString = Resources.toString(url, StandardCharsets.UTF_8);
-      final CombinedConnectorCatalog catalog = Jsons.deserialize(jsonString, CombinedConnectorCatalog.class);
-      return catalog;
+      LOGGER.info("Loading local catalog definitions from {}", url);
```
pedroslopez, line 37: nit: add the word `connector` in there

pedroslopez: also i'm not sure how often this will be logged - should probably only happen on initialization if possible

pedroslopez (approved):

airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
```diff
@@ -14,13 +14,8 @@ public class CatalogDefinitionsConfig {
...
    */
   public static String getLocalConnectorCatalogPath() {
-    Optional<String> customCatalogPath = new EnvConfigs().getLocalCatalogPath();
-    if (customCatalogPath.isPresent()) {
-      return customCatalogPath.get();
-    }
-
-    return AirbyteCatalogConstants.DEFAULT_LOCAL_CONNECTOR_CATALOG_PATH;
-
+    final Optional<String> customCatalogPath = new EnvConfigs().getLocalCatalogPath();
```
pedroslopez, line 17: nit: this should have used micronaut configs and specified the env var via the application.yml, instead of EnvConfigs, which will go away soon.

docker-compose.yaml
```diff
@@ -28,6 +28,7 @@ services:
       - DATABASE_URL=${DATABASE_URL}
       - DATABASE_USER=${DATABASE_USER}
       - LOG_LEVEL=${LOG_LEVEL}
+      - LOCAL_CONNECTOR_CATALOG_PATH=${LOCAL_CONNECTOR_CATALOG_PATH}
```
pedroslopez, line 31: Do we need to add this to any of the other apps? I think the bootloader is the only place where this is used because that's where we seed things, but just checking in case we use it in the server or elsewhere as well for some reason (i believe the server should just be using whatever is in the db rather than the local provider)

evantahler (approved):
👍 from me on the DX, but I defer my technical review to the others.

docker-compose.yaml
bnchrch: I tested around and it seems like its only the boot loader that requires it

prateekmukhedkar:
@bnchrch fetched this PR locally, to confirm I need to add a new key to `.env.dev` here and set it the full path of source definitions? (or folder name?)

`LOCAL_CONNECTOR_CATALOG_PATH=/Users/prateekmukhedkar/code/airbyte/airbyte-config/init/src/main/re`sources/seed/source_definitions.yaml`

bnchrch:
@prateekmukhedkar Would you be able to hold off on using this for a couple more hours.

Im running this again this morning and something very strange is going on with the environment variables in docker.

I just want to make sure this will 100% work for you.

bnchrch:
Sorry to be a blocker.

but here are a few steps that should unblock yourself if you need to 

## 1. Generating the OSS Catalog
In the `airbyte` repo run
```sh
SUB_BUILD=ALL_CONNECTORS ./gradlew :airbyte-config:specs:generateOssConnectorCatalog
```

## 2. Move the OSS catalog to platform
In the `airbyte` repo run
```sh
cp airbyte-config/init/src/main/resources/seed/oss_catalog.json <PATH_TO_PLATFORM_REPO>/airbyte-config/init/src/main/resources/seed/oss_catalog.json
```

### 3. Disable catalog download
Comment out these lines in airbyte-platform
https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-config/specs/build.gradle#L15-L45

### 4. Run platform
In the `airbyte-platform` run
```sh
VERSION=dev docker-compose up
```

bnchrch:
closed in favor of https://github.com/airbytehq/airbyte-platform-internal/pull/4705

PR #133, opened 2023-02-24
🪟🐛 Connectors form: Handle errors for string/number arrays better

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/23329 , making sure errors are correctly shown for array inputs (showing multiple errors one below the other)

Screenshot 2023-02-24 at 12 17 45

## How

There are two parts to this:

* When changing the value of a tag input and the field is not touched yet, touch it manually (because it won't happen otherwise and errors are not shown) - `airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/Control.tsx`
* Extend property section to handle arrays of errors - `airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/PropertySection.tsx`
* If the items type of a field in the schema has a pattern defined, copy it into the form block so it can be used in the property section to correctly render the pattern message - `airbyte-webapp/src/core/form/schemaToFormBlock.ts`


airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/PropertySection.tsx
```diff
@@ -47,13 +62,20 @@ export const PropertySection: React.FC<PropertySectionProps> = ({ property, path
 
   const hasError = !!meta.error && meta.touched;
 
-  const errorValues = meta.error === "form.pattern.error" ? { pattern: property.pattern } : undefined;
-  const errorMessage = <FormattedMessage id={meta.error} values={errorValues} />;
+  const errorMessage = Array.isArray(meta.error) ? (
+    <FlexContainer direction="column" gap="none">
+      {meta.error.filter(Boolean).map((error, index) => {
+        return <ErrorMessage key={index} error={error} property={property} />;
+      })}
```
lmossman, line 69: I don't think it is very useful to repeat the same error message multiple times, but this is good for showing multiple errors if they are different.

Could we unique the error messages here so there aren't duplicates?

lmossman (approved):
Works as expected locally and changes look good.

Had one comment, probably doesn't need another look once addressed so I'm going to approve

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/PropertySection.tsx
flash1293: Good point, I'm going to adjust this

PR #136, opened 2023-02-24
🪟🐛 Connector builder: Show page size option for cursor strategy

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/23350

With this fix, the page size option for the paginator cursor strategy should be shown if there is a value set for the page size

## How

The problem is described very well here: https://github.com/airbytehq/airbyte/issues/23350#issuecomment-1442570483

This PR fixes this by mapping the fastfield for the oneOf block to the whole object instead of just its type property. This causes some additional render calls on some components because every change inside of the oneOf block will re-render the whole component, but it makes sure that the passed in children will be up-to-date with the latest changes if they are conditional (like in the linked issue).

Testing with CPU throttling, the UI doesn't feel any slower than before using this approach.


lmossman:
I pushed up one [small change](https://github.com/airbytehq/airbyte-platform/pull/136/commits/353b67872ab37f30566d7e39017c28b3de7841bc) here to fix another issue that I noticed when testing this locally:

if you typed `0` into Page size, then a `0` would be rendered below the Page size input. This is because we were doing `pageSizeField.value && ...` for the conditional rendering, so in the case that it was 0 it wouldn't evaluate the rest of the && chain and instead just render a `0` like so:
Screen Shot 2023-02-24 at 11 49 54 AM

Fixed by changing that condition to be `pageSizeField.value !== undefined && ...`

lmossman (approved):
The fix makes sense, tested locally and it is now working as expected. LGTM

PR #137, opened 2023-02-24
🪟 🐛 Fix docs download

timroes:
## What

Fixes the doc download script. This currently copied `docs` **into** `./docs` (i.e. created a `./docs/docs` folder) and thus had a wrong nesting layer.

## How

Changing this to only copy the actual 2 folder we need (and also making sure they land in the right pathes). Also validate that a file we know should exist actually exist in the right path, otherwise fail the build.

I validated that the docs are bundled by checking inside the `airbyte/webapp:dev-c42ae05a14` image that has been published by this task, that it contains the docs under `/usr/share/nginx/html/docs`.


josephkmh (approved):

cpdeethree (approved):
Thanks for fixing this! not sure how I was able to see docs in dev-3 originally when I tested this. Maybe they were cached or something.

PR #138, opened 2023-02-24
Connector builder: Load and update project

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/22446 by loading and saving the manifest via the API.

When navigating to `/workspaces/<workspace id>/connector-builder/edit/<project id>`, the manifest of the project is loaded and displayed. When editing, the manifest is saved back after there weren't any changes for 2.5 seconds.

### All saved
Screenshot 2023-02-27 at 18 44 05
Screenshot 2023-02-28 at 10 18 18

### Saving in progress (or waiting for 2.5 second debounce)
Screenshot 2023-02-27 at 18 44 13
Screenshot 2023-02-28 at 10 18 23

### Draft can't be saved because the form is invalid
Screenshot 2023-02-27 at 18 48 02
Screenshot 2023-02-28 at 10 18 31

### Modal shown when trying to navigate away while the form is invalid
Screenshot 2023-02-27 at 18 47 57

### Modal shown when trying to navigate away while saving is in progress (will auto-close once saving is done):
Screenshot 2023-02-27 at 19 26 41

### Error message shown when the update request to the API failed (should be rare)
Screenshot 2023-02-27 at 18 48 35
Screenshot 2023-02-28 at 10 18 47

### `services/connectorBuilder/ConnectorBuilderApiService.ts`

Add a hook to use the resolve endpoint via suspense query in order to use it as part of the initialization routine when opening a builder project

### `services/connectorBuilder/ConnectorBuilderProjectsService.ts`

Adding hooks to fetch the project details (plus manifest) and a mutation query for updating it

### `services/connectorBuilder/ConnectorBuilderLocalStorageService.tsx`

Removing all parts except for remembering whether the user prefers ui or yaml (handled by the state service now)

### `services/connectorBuilder/ConnectorBuilderStateService.tsx`

This is the main part of the PR.

Instead of loading manifest and form values from local storage, do the following:
* Get projectId from route
* Execute `useInitializedBuilderProject` sub-hook
  * Load project via the suspense query
  * Resolve manifest via resolve endpoint via suspense query - this step might fail if the manifest can't be resolved - the error is swallowed and no resolved manifest is returned (can still use the yaml editor in this case)
  * If there is a resolved manifest, try to convert into form values
  * If it works, use that as the initial form values
  * If not or if the manifest couldn't be resolved, go with the default form values (but inject the real connector name as it's also relevant for the yaml view)
* Use a regular state hook for the manifest and the form values instead of a localstorage-backed one
* If the form values conversion couldn't be performaned, switch to the yaml view (only initially, later calls are managed by the click handler)
* Remove the ref that was only relevant because the localstorage-backed state hook would always serialize/deserialize the passed in object
* Track the persisted state (manifest and name) to determine whether there are local changes not saved yet
* `getSavingState` determines the current state of saving:
  * `saved` if the persisted state matches the current state
  * `invalid` if the current view is invalid and thus can't be saved
  * `loading` if the current view is valid and the persisted state does not match the current state
  * `error` if the mutation query hook returned an error on last run
* Use an effect hook to trigger a call to the mutation query hook with the current manifest and connector name to actually update - once that succeeds, set the persisted state so the saving state is updated
* Make sure no local changes are lost in `useBlockOnSavingState` sub hook
  * Block navigation if saving state is not `saved`
  * If it's invalid, show a confirmation modal explaining what's going on
  * If it's loading, show a spinner modal which auto-retries the navigation once the saving state switches again
* Export saving state and `triggerUpdate` via context so it can be consumed by the `SavingIndicator` component

### `components/connectorBuilder/Builder/SavingIndicator.tsx`

Component used to show the current saving state. Consumes the newly exported things from the context. Used for both form and yaml view.

As this requires a super small spinner, I changed the global `Spinner` component from a `small` prop to a `size` prop that takes `xs`, `sm` and `md` instead (had some follow-up changes across the webapp as this component is used in a bunch of places)

### `components/connectorBuilder/Builder/BuilderSidebar.tsx`

Took the opportunity to get rid of a bunch of hard to maintain spacing-related css rules used in the builder sidebar and use `FlexContainer`s instead. Also to make the positioning simpler I introduced global variables for the size of toggle button and side bar to position it the same way in both form and yaml mode in a stable way.

### `components/connectorBuilder/YamlEditor/YamlEditor.tsx`

Besides rearranging the layout a bit to show the saving indicator, avoid setting the manifest on editor load to avoid an update call immediately after loading the UI.

### `components/connectorBuilder/useManifestToBuilderForm.ts`

Split out the internal `resolve` call in a separate function so we can call `convertToBuilderFormValuesSync` synchronously to use it as part of the `useInitializedBuilderProject` sub hook. Doing this I also changed the interface to not take in the whole `currentBuilderFormValues` but only the name (as this is what the API returns) which requires some adjustments across the builder code.

### `components/connectorBuilder/types.ts`

This is a small fix - on the low code beta PR I thought version 0.28 was the cutover, but it's actually 0.29

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨


flash1293:
@lmossman this is not fully functional yet because the create part is missing to be able to fix all the errors on the create page and to have an e2e-testable state, but the code changes can be reviewed already on this branch.

To use it, you need to "seed" a builder project with a manifest by calling the create API manually (e.g. curl or postman):
```
POST {{baseUrl}}/v1/connector_builder_projects/create
{
    "workspaceId": "e1ee1ca1-db59-4fc7-ab63-17b41cdb41f2",
    "builderProject": {
        "name": "Whisky Hunter",
        "draftManifest": {
            "version": "0.1.0",
            "definitions": {
                "selector": {
                    "extractor": {
                        "field_path": []
                    }
                },
                "requester": {
                    "url_base": "https://whiskyhunter.net/api",
                    "http_method": "GET"
                },
                "retriever": {
                    "record_selector": {
                        "$ref": "#/definitions/selector"
                    },
                    "paginator": {
                        "type": "NoPagination"
                    },
                    "requester": {
                        "$ref": "#/definitions/requester"
                    }
                },
                "base_stream": {
                    "retriever": {
                        "$ref": "#/definitions/retriever"
                    }
                },
                "auctions_data_stream": {
                    "$ref": "#/definitions/base_stream",
                    "$parameters": {
                        "name": "auctions_data",
                        "path": "/auctions_data/?format=json"
                    }
                },
                "auctions_info_stream": {
                    "$ref": "#/definitions/base_stream",
                    "$parameters": {
                        "name": "auctions_info",
                        "path": "/auctions_info?format=json"
                    }
                },
                "distilleries_info_stream": {
                    "$ref": "#/definitions/base_stream",
                    "$parameters": {
                        "name": "distilleries_info",
                        "path": "/distilleries_info/?format=json"
                    }
                }
            },
            "streams": [
                "#/definitions/auctions_data_stream",
                "#/definitions/auctions_info_stream",
                "#/definitions/distilleries_info_stream"
            ],
            "check": {
                "stream_names": [
                    "auctions_data",
                    "auctions_info",
                    "distilleries_info"
                ]
            }
        }
    }
}
```

then navigate to `https://127.0.0.1:3000/workspaces/<your workspace id>/connector-builder/edit/<the returned project id>`

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
```diff
@@ -66,97 +59,87 @@ interface BuilderSidebarProps {
...
-
-        <AddStreamButton onAddStream={(addedStreamNum) => handleViewSelect(addedStreamNum)} data-testid="add-stream" />
-      </div>
+        <div className={styles.connectorName}>
+          <Heading as="h2" size="sm">
+            {values.global?.connectorName}
+          </Heading>
+        </div>
+
+        <FlexItem>
```
lmossman, line 89: It doesn't seem like wrapping the `<SavingIndicator>` in a `<FlexItem>` is necessary here, since you aren't passing any props to the FlexItem

airbyte-webapp/src/components/connectorBuilder/Builder/SavingIndicator.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return (
+      <FlexContainer gap="sm" alignItems="center">
+        <Spinner size="xs" />
+        <FormattedMessage id="connectorBuilder.loadingState.loading" />
+      </FlexContainer>
+    );
+  }
+  if (savingState === "error") {
+    return (
+      <Button variant="clear" onClick={triggerUpdate}>
```
lmossman, line 41: UX nit: I tested this state out by shutting down my local docker compose deployment, then making a change in the builder UI. In this state, when I clicked this button to retry, there was no UX feedback for me that anything happened, likely because the request failed so quickly that the `loading` state didn't have time to be visible. 

This probably isn't worth spending much time on, but maybe we could force this to stay in `loading` for some amount of time, say 200ms, so that the user at least sees that something was tried. What do you think?

airbyte-webapp/src/components/connectorBuilder/YamlEditor/YamlEditor.tsx
```diff
@@ -54,7 +57,12 @@ export const YamlEditor: React.FC<YamlEditorProps> = ({ toggleYamlEditor }) => {
       try {
         const json = load(yamlValue) as ConnectorManifest;
         setYamlIsValid(true);
-        debouncedSetJsonManifest(json);
+        // skip setting the manifest on the first load as it just got passed in and is synced already
+        if (initialLoad.current) {
+          initialLoad.current = false;
```
lmossman, line 62: Should we have a similar thing for the builder form when switching from YAML -> UI?

It seems like currently it always calls `/update` whenever I switch from YAML to UI, even if the manifest is unchanged as a result of switching

airbyte-webapp/src/components/connectorBuilder/useManifestToBuilderForm.test.ts
```diff
@@ -57,7 +56,7 @@ describe("Conversion throws error when", () => {
       throw new Error(errorMessage);
     };
     const convert = async () => {
-      return convertToBuilderFormValues(resolve, {} as ConnectorManifest, DEFAULT_BUILDER_FORM_VALUES);
+      return convertToBuilderFormValues(resolve, {} as ConnectorManifest, "Untitled");
```
lmossman, line 59: Could we put `"Untitled"` into a constant somewhere and reference that constant in all of these tests instead of repeating the string? This will make it easier to change in the future if we want

lmossman (commented):
Had a bunch of comments, but overall this is looking and working great. Lmk if you'd like me to take another look after you address

PR #140, opened 2023-02-24
Cleanup docs folder

timroes:
## What

Remove the `docs` folder and add it to `.gitignore`. It has accidentally be added as part of https://github.com/airbytehq/airbyte-platform/pull/69/files, because I assume this was started before https://github.com/airbytehq/airbyte-platform/pull/66 and merge conflicts where resolved wrongly.

**This should be merged after https://github.com/airbytehq/airbyte-platform/pull/137**

## Can this PR be safely reverted / rolled back?
*If you know that your PR is backwards-compatible and can be simply reverted or rolled back, check the YES box.*

*Otherwise if your PR has a breaking change, like a database migration for example, check the NO box.*

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

PR #141, opened 2023-02-24
Create source with secret

JonsSpaghetti:
## What
Allows creating a source with a secret ID without affecting the current codepath.

## How
Adds an optional param to the SourceCreate request body which (if present) rehydrates the secret, uses the connector specification to determine where to put secrets in the passed in connectionConfiguration, and puts them there.

If not present, we keep the current behavior.

Tested locally using google sheets (authSpecification) and github OAuth (advanced_auth) in order to create a source.

There are a few limitations that we're going to call acceptable for the config API and will be handled differently for the public API PR.
i.e. missing the oneOf discriminator for credentials blocks == return an error (in the public API we'll handle that for users but for config API users, the form should theoretically always pass us a value there).

## Recommended reading order
1. `SourceHandler.java`
2. `OAuthSecretHelper.java`
3. Others

*If unsure, leave it blank.*
- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
There should be no breaking changes.


airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceHandler.java
```diff
@@ -367,4 +381,11 @@ protected static SourceSnippetRead toSourceSnippetRead(final SourceConnection so
         .icon(SourceDefinitionsHandler.loadIcon(sourceDefinition.getIcon()));
   }
 
+  @VisibleForTesting
+  JsonNode hydrateOAuthResponseSecret(String secretId) {
+    final SecretCoordinate secretCoordinate = SecretCoordinate.fromFullCoordinate(secretId);
+    return secretsRepositoryReader.fetchSecret(secretCoordinate);
+
+  }
```
JonsSpaghetti, line 389: I made this a function because it makes it slightly easier for me to test

airbyte-commons/src/main/java/io/airbyte/commons/json/Jsons.java
```diff
@@ -608,4 +608,60 @@ public DefaultPrettyPrinter withSeparators(final Separators separators) {
...
+        currentNode.set(key, node);
+      }
+    }
+    return node;
+  }
+
+  public static void setNestedValue(final JsonNode json, final List<String> keys, final JsonNode replacement) {
+    setNested(json, keys, (node, finalKey) -> node.set(finalKey, replacement));
+  }
+
```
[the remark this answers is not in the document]
JonsSpaghetti: Oh I suppose merge Nodes is separate from the setNestedValue changes. Used to allow us to merge nodes same as the frontend does when handling merging legacy authSpecification complete_oauth responses into the passed in connectionConfiguration.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/helpers/OAuthSecretHelper.java
```diff
@@ -0,0 +1,100 @@
...
+        final List<String> jsonPathList = entry.getValue();
+
+        Jsons.setNestedValue(newConnectionConfiguration, jsonPathList, hydratedSecret.get(key));
+      }
+      return newConnectionConfiguration;
+    } else {
+      // Just merge, complete_oauth handled setting rootNode for us.
+      return Jsons.mergeNodes(connectionConfiguration, hydratedSecret);
+    }
+  }
```
JonsSpaghetti, line 52: @flash1293 a quick review here to make sure we're matching up with what frontend does would be awesome.

JonsSpaghetti:
Removal of generated docs folder + addition to the gitignore is per Tim's PR: https://github.com/airbytehq/airbyte-platform/pull/140

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/helpers/OAuthSecretHelper.java
flash1293: Went through the code and from what I can tell it encodes the same logic like https://github.com/airbytehq/airbyte-cloud/blob/873c4ae061b1efda14f31265d50323b4be42b9a5/oss/airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/auth/useOauthFlowAdapter.tsx#L33

Did not test locally.

PR #143, opened 2023-02-24
🪟 🧹 Provide clearer messaging if a dbt Cloud account has no associated jobs

tealjulia:
## What
Related to https://github.com/airbytehq/oncall/issues/1562

Previously, if a dbt Cloud account didn't have any jobs, the Airbyte UI would render a pretty meaningless empty dropdown when users clicked "add transformation".

Now, we render explanatory text if there are no jobs found.

BEFORE:
An empty dropdown? Hard to tell what's going on.
Screenshot 2023-02-24 at 3 12 24 PM

AFTER:
Specific message, no deceptive unusable button
Screenshot 2023-02-24 at 3 11 53 PM

## How
Conditionally render the dropdown

*If unsure, leave it blank.*
- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
Users will not see the dropdown that is unusable, and will instead see a more useful message


airbyte-webapp/src/pages/connections/ConnectionTransformationPage/DbtCloudTransformationsCard/DbtJobsForm.tsx
```diff
@@ -62,9 +62,14 @@ export const DbtJobsForm: React.FC<DbtJobsFormProps> = ({
                       <span className={styles.cardTitle}>
                         <FormattedMessage id="connection.dbtCloudJobs.cardTitle" />
                         <DropdownMenu
-                          options={availableDbtCloudJobs
-                            .filter((remoteJob) => !values.jobs.some((savedJob) => isSameJob(remoteJob, savedJob)))
-                            .map((job) => ({ displayName: job.jobName, value: job }))}
+                          options={
```
ambirdsall, line 65: Some thoughts here:
1) we have to put the logic which detects an empty job list upstream of the `.filter((remoteJob) => !values.jobs.some((savedJob) => isSameJob(remoteJob, savedJob)))`, so we don't show a "we can't find any jobs!" message to users when we've found the jobs just fine but filtered them all out
1) with the extra logic, I think we've hit the point where we need to pull the processed dropdown options to their own variable instead of doing it all inline.

airbyte-webapp/src/packages/cloud/locales/en.json
```diff
@@ -65,6 +65,7 @@
   "connection.dbtCloudJobs.genericError": "There was an error communicating with dbt Cloud.",
   "connection.dbtCloudJobs.explanation": "After an Airbyte sync job has completed, the following jobs will run",
   "connection.dbtCloudJobs.noJobs": "No transformations",
+  "connection.dbtCloudJobs.noJobsFoundForAccount": "No transformations found for this dbt Cloud account",
```
ambirdsall, line 68: Maybe "no jobs found for this dbt Cloud account"? Seems like we should use dbt's terminology here, because that's what users will likely need to cross-reference to fix issues.

airbyte-webapp/src/pages/connections/ConnectionTransformationPage/DbtCloudTransformationsCard/DbtJobsForm.tsx
tealjulia: Mostly agree!

I think we can just conditionally render the dropdown/show some messaging if no jobs were found.

tealjulia:
My main misgiving with this is that it isn't particularly actionable, but I'm not sure where to direct folks!  

I don't think [our docs](https://docs.airbyte.com/cloud/dbt-cloud-integration/) address this.  We generally don't link external docs in the product, but there [are these](https://docs.getdbt.com/docs/get-started/getting-started/building-your-first-project/schedule-a-job#create-and-run-a-job) on how to create a dbt job.  I don't love telling a user they're in a hole without giving them a shovel, but even as is it is more useful than what we had before.

ambirdsall:
> this... isn't particularly actionable, but I'm not sure where to direct folks

@tealjulia maybe we could give a link to https://cloud.getdbt.com/? That's where the fix would have to happen.

tealjulia:
migrated

PR #144, opened 2023-02-24
Return FailureReason as part of SynchronousJobRead

erohmensing:
## What
* close https://github.com/airbytehq/airbyte/issues/18514
* Return `FailureReason` in API response as part of the `jobInfo` of synchronous jobs
* Add testing for `jobCoverter.getSynchronousJobRead`

Profit: FailureReason populates in the API, if the connector outputs a trace message! 
image

## How
* Add `failureReason` from `ConnectorJobOutput` to `SynchronousJobMetadata`
* Add `failureReason` to `SynchronousJobRead`, which is used in populating the API response for synchronous jobs

## Recommended reading order
1. https://github.com/airbytehq/airbyte-platform/blob/7400ce667b93a7970dd58d5a47e47d4030abaa7a/airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/SynchronousResponse.java#L48
2. https://github.com/airbytehq/airbyte-platform/blob/7400ce667b93a7970dd58d5a47e47d4030abaa7a/airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/SynchronousJobMetadata.java#L49
3. https://github.com/airbytehq/airbyte-platform/blob/7400ce667b93a7970dd58d5a47e47d4030abaa7a/airbyte-commons-server/src/main/java/io/airbyte/commons/server/converters/JobConverter.java#L248
4. https://github.com/airbytehq/airbyte-platform/blob/7400ce667b93a7970dd58d5a47e47d4030abaa7a/airbyte-api/src/main/openapi/config.yaml#L4270
5. https://github.com/airbytehq/airbyte-platform/blob/7400ce667b93a7970dd58d5a47e47d4030abaa7a/airbyte-commons-server/src/test/java/io/airbyte/commons/server/converters/JobConverterTest.java#L291

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
None


airbyte-commons-server/src/test/java/io/airbyte/commons/server/converters/JobConverterTest.java
```diff
@@ -60,226 +46,276 @@
...
+    assertTrue(Enums.isCompatible(FailureReason.FailureOrigin.class, io.airbyte.api.model.generated.FailureOrigin.class));
 
   }
 
-  @Test
-  void testGetJobInfoRead() {
-    assertEquals(JOB_INFO, jobConverter.getJobInfoRead(job));
-  }
+  @Nested
+  class TestJob {
```
erohmensing, line 77: ⚠️ The diff for this file is very messy, because I took the all but one of the existing tests and tested them into `TestJob` so that I could also create a new `TestSynchronousJob`, since the entire `setup()` was different. 

TLDR of changes: 
* testEnumConversion() is a base level test
* all other old tests into `TestJob`, all constants except for ones shared with `TestSynchronousjob` also go under `TestJob`
* new test in `TestSynchronousJob`

+ new test, no changes to old tests.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SchedulerHandler.java
```diff
@@ -472,6 +472,8 @@ private CheckConnectionRead reportConnectionStatus(final SynchronousResponse<Sta
           .status(Enums.convertTo(response.getOutput().getStatus(), StatusEnum.class))
           .message(response.getOutput().getMessage());
     } else {
+      // TODO (Ella): Update logic - 'status' should no longer be necessary, as is the case for the other
+      // two synchronous jobs.
```
erohmensing, line 476: Creating a separate issue for this change, one I didn't realize was desired during the tech spec, but I think makes sense. Will attach here. TL;DR a lack of output should indicate that the check job itself (not checking the connection) failed.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/converters/JobConverter.java
```diff
@@ -235,6 +227,20 @@ public LogRead getLogRead(final Path logPath) {
     }
   }
 
+  public static FailureReason getFailureReason(final io.airbyte.config.FailureReason failureReason) {
```
pedroslopez, line 230: We should mark the `failureReason` arg as `@Nullable` here

pedroslopez (approved):
Nice work! Code LGTM

evantahler (commented):
Nice work! 
I don't feel qualified to approve this work personally, and I'll add someone from the Platform team to review as well.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/SynchronousResponse.java
```diff
@@ -61,7 +61,8 @@ public static <T, U> SynchronousResponse<T> fromTemporalResponse(final TemporalR
         configId,
         jobOutput != null ? jobOutput.getConnectorConfigurationUpdated() : false,
         createdAt,
-        endedAt);
+        endedAt,
+        jobOutput != null ? jobOutput.getFailureReason() : null);
```
pedroslopez, line 65: Since we're using a few things from the jobOutput now, I wonder if it makes sense just to pass the jobOutput through and let metadata pull things from it as needed

erohmensing: That makes sense to me, i'll try it out

erohmensing: Any idea why we're passing `jobOutput` and `responseOutput` separately to this method, when they both came from the `temporalResponse` in the first place? That might be a bit of cleanup as well

erohmensing: @pedroslopez lmk what you think of the result. I think it looks cleaner - a little less DRY [here](https://github.com/airbytehq/airbyte-platform/pull/144/files#diff-754353c954321b0adc883a661e79b4ae6e961e3965a0059ab26e691647697751R56-R57) but with a bit of a refactor of `track()` in [this issue ](https://app.zenhub.com/workspaces/connector-operations-6318ede5eda32e0010ef8566/issues/gh/airbytehq/airbyte/19898) we'll be wanting to send the full `jobOutput` there too, so I think we'll stop pulling those out in `execute` anyway

airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/SynchronousJobMetadata.java
```diff
@@ -29,24 +32,26 @@ public class SynchronousJobMetadata {
...
    * @param configType job type
    * @param configId id of resource for job type (i.e. if configType is discover config id is going to
    *        be a source id)
-   * @param connectorConfigurationUpdated whether this job updated the connector configuration
    * @param createdAt time the job was created
    * @param endedAt time the job ended
    * @return synchronous job metadata
    */
   public static SynchronousJobMetadata fromJobMetadata(final JobMetadata jobMetadata,
+                                                       @Nullable ConnectorJobOutput jobOutput,
```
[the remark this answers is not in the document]
erohmensing: Thanks for the catch! I honestly thought we had CI that would yell at me for this 😄

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SchedulerHandler.java
erohmensing: https://app.zenhub.com/workspaces/connector-operations-6318ede5eda32e0010ef8566/issues/gh/airbytehq/airbyte/23451

erohmensing:
Ported to https://github.com/airbytehq/airbyte-platform/pull/144!

PR #145, opened 2023-02-24
Stream status on connection list page

krishnaglick:
## What
Resolves https://github.com/airbytehq/airbyte/issues/23342
Adds a new cell to the connection list to show stream status
To test: Add `"connection.streamCentricUI.v2": true` to your `airbyte-webapp/.experiments.json`

## How
Since we don't have per-stream status I'm merging in some of the connection status to fake it so ideally when the backend starts to provide it we don't need to do much leg work.

## Notes
1. I'm not sure how best to test this. We can confirm that the numbers but since the feature is mostly visual I feel it's difficult to test. With that said, since it's visual any regressions should be very visible.
2. The StreamStatusCell is a very dense component. I'm planning to reuse some of this logic on the actual stream status page/tab so I'll be pulling out pieces as necessary in a later PR.
3. This is PR'd into the initial feature flag and status page work. That work is ready to merge so I expect no issues from this.


airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { Checkmark } from "./StreamStatusIcons/Checkmark";
+import { Error } from "./StreamStatusIcons/Error";
+import { Inactive } from "./StreamStatusIcons/Inactive";
+import { Late } from "./StreamStatusIcons/Late";
+import { Syncing } from "./StreamStatusIcons/Syncing";
+import { AirbyteStreamAndConfiguration } from "../../../core/request/AirbyteClient";
+import { ConnectionTableDataItem } from "../types";
+
+// TODO: When we have Actionable Errors, uncomment
+type StatusType = /* "Action Required" | */ "On Track" | "Disabled" | "Error" | "Behind";
```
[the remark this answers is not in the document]
krishnaglick: I know there was some discussion about enum vs types since the strings are type-enforced. I'm not the biggest fan of enums but curious what @edmundito thinks.

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
```diff
@@ -73,30 +76,11 @@ const ConnectionTable: React.FC<ConnectionTableProps> = ({ data, entity, onClick
...
-            status={props.row.original.lastSyncStatus}
-            value={props.cell.getValue()}
-            enabled={props.row.original.enabled}
-          />
-        ),
-      }),
+  const columns = React.useMemo(() => {
+    // ColumnDef requires explicit typing to unshift down below, but each column has explicit
+    // typing so we need an `any` to accommodate for varied column types
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
dizel852: + We don't need manually control what columns we need to pass, react-table will handle it under the hood. 
+ Also, in the future, if we will have one more column controlled by a feature flag, that would be easier to implement.
+ Less code - one more column vs if/else checks

airbyte-webapp/src/components/EntityTable/components/StreamStatusIcons/Checkmark.tsx
```diff
@@ -0,0 +1,9 @@
+export const Checkmark = () => (
```
edmundito, line 1: Should these icons go in `src/components/icons`? Is there anything special that makes them unique to this component?

airbyte-webapp/src/components/EntityTable/components/StreamStatusIcons/Checkmark.tsx
```diff
@@ -0,0 +1,9 @@
...
+  <svg width="30" height="30" viewBox="0 0 30 30" fill="none" xmlns="http://www.w3.org/2000/svg">
```
edmundito, line 2: I always assumed that this xmlns="http://www.w3.org/2000/svg" is removed when the svg is embedded into html.

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
+import { CellContext } from "@tanstack/react-table";
```
edmundito, line 1: EntityTable is a bit overloaded, and this component should be standalone moved to `components/connection/StreamStatusCell` so that it can be easily broken down into multiple files.

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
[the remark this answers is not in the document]
edmundito: After looking at this, I would argue for an enum (const enum specifically) since these statuses are representative of the data and as @dizel852 pointed out in another comment it would be easier to track. The values should be camelCase (or `snake_case` if we expect these to come from the backend int he future) but one advantage of camelCase is that they can be applied to i18n strings more easily.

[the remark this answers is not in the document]
edmundito: I would also name this `StreamStatus` or `StreamStatusType` to make search easier.

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+  streams.reduce(
+    (sortedStreams, stream) => {
+      sortedStreams[getStatusForStream(stream)]++;
+      return sortedStreams;
+    },
+    // This is the intended display order thanks to Javascript object insertion order!
+    { /* "Action Required": 0, */ Error: 0, Behind: 0, "On Track": 0, Disabled: 0 }
+  );
+
+const StreamsBar: React.FC<{ streams: AirbyteStreamWithStatusAndConfiguration[] }> = ({ streams }) => {
```
edmundito, line 121: I imagine that this streams bar can be shared with the actual status page? Maybe it should be its own separate component?

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const sortedStreams = sortStreams(streams);
+  return (
+    <div className={styles.bar}>
+      {(Object.entries(sortedStreams) as Array<[StatusType, number]>)
+        .filter(([, count]) => !!count)
```
edmundito, line 126: Might be cleaner to extract this value into a var first

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+        // This can be undefined due to historical data, but should always be present
+        stream.config.scheduleType &&
+        !["cron", "manual"].includes(stream.config.scheduleType) &&
+        stream.config.latestSyncJobCreatedAt &&
+        stream.config.scheduleData?.basicSchedule?.units &&
+        stream.config.latestSyncJobCreatedAt * 1000 < // x1000 for a JS datetime
+          dayjs()
+            // Subtract 2x the scheduled interval and compare it to last sync time
+            .subtract(stream.config.scheduleData.basicSchedule.units, stream.config.scheduleData.basicSchedule.timeUnit)
+            .valueOf()
```
edmundito, line 99: Maybe extract this into `isStreamBehind(stream): boolean`?

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+                {count} <Syncing className={styles.syncing} />
+              </div>
+            ) : null}
+          </div>
+        ))}
+    </>
+  );
+};
+
+const StreamStatusCellTooltipContent = ({ streams }: { streams: AirbyteStreamWithStatusAndConfiguration[] }) => {
```
edmundito, line 160: Semantics... but even though it's using a tooltip to render the content, this is more of a Popover component since the function of the tooltip is to explain something. `StreamStatusPopover`?

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+} from "core/request/AirbyteClient";
+import { useGetConnection } from "hooks/services/useConnectionHook";
+
+import styles from "./StreamStatusCell.module.scss";
+import { Checkmark } from "./StreamStatusIcons/Checkmark";
+import { Error } from "./StreamStatusIcons/Error";
+import { Inactive } from "./StreamStatusIcons/Inactive";
+import { Late } from "./StreamStatusIcons/Late";
+import { Syncing } from "./StreamStatusIcons/Syncing";
+import { AirbyteStreamAndConfiguration } from "../../../core/request/AirbyteClient";
```
edmundito, line 23: Relative import

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+const StreamsPerStatus: React.FC<{ streams: AirbyteStreamWithStatusAndConfiguration[] }> = ({ streams }) => {
+  const sortedStreams = sortStreams(streams);
+  return (
+    <>
+      {(Object.entries(sortedStreams) as Array<[StatusType, number]>)
+        .filter(([, count]) => !!count)
+        .map(([statusType, count], index) => (
+          <div className={styles.tooltipContent} key={statusType}>
+            <div className={styles.streamsDetail}>
+              {iconMap[statusType]} <b>{count}</b> {statusType}
```
edmundito, line 147: statusType here would need to be resolved from `en.json`

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
```diff
@@ -189,9 +173,44 @@ const ConnectionTable: React.FC<ConnectionTableProps> = ({ data, entity, onClick
...
-    ],
-    [columnHelper, sortBy, sortOrder, onSortClick, entity, allowAutoDetectSchema]
-  );
+    ];
+
+    if (streamCentricUIEnabled) {
+      columns.unshift(
+        columnHelper.display({
+          id: "stream-status",
+          cell: StreamsStatusCell,
```
edmundito, line 182: The cell doesn't have a fixed width, so it gets longer as the table resizes

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.module.scss
```diff
@@ -0,0 +1,84 @@
...
+      background: colors.$red;
+    }
+
+    &.disabled {
+      background: colors.$grey;
+    }
+  }
+}
+
+$contentWidth: 75%;
```
edmundito, line 36: Instead of using a fixed width, it's better to use horizontal padding to push the content inward.

edmundito (commented):
Really great to see this up and running very quickly!

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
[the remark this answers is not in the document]
krishnaglick: I'm sorry, I mean what changes would need to take place within the NextTable component to enable this functionality. I'm all for it.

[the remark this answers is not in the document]
dizel852: Oh , here they are : 

```
export interface TableProps<T> {
  className?: string;
  .........
  testId?: string;
  columnVisibility?: VisibilityState;   <====
}
```
```
export const NextTable = <T,>({
  testId,
  className,
 .........
  columnVisibility,    <====
}: PropsWithChildren<TableProps<T>>) => {
.......
```

```
const table = useReactTable({
    columns,
    data,
    initialState: {
      columnVisibility,    <====
    },
    getCoreRowModel: getCoreRowModel(),
  });
```

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
krishnaglick: Some of this logic is going to be shared with the stream-centric status page. I was planning on extracting that out then.

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
krishnaglick: Yes, I am planning to split this out when I start that work.

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
krishnaglick: sure!

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
krishnaglick: Added `size: 170,`

airbyte-webapp/src/components/EntityTable/components/StreamStatusIcons/Checkmark.tsx
krishnaglick: Nope, I'll move.

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { Late } from "./StreamStatusIcons/Late";
+import { Syncing } from "./StreamStatusIcons/Syncing";
+import { AirbyteStreamAndConfiguration } from "../../../core/request/AirbyteClient";
+import { ConnectionTableDataItem } from "../types";
+
+// TODO: When we have Actionable Errors, uncomment
+type StatusType = /* "Action Required" | */ "On Track" | "Disabled" | "Error" | "Behind";
+
+const statusMap: Readonly<Record<StatusType, string>> = {
+  "On Track": styles.onTrack,
```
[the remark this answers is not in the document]
krishnaglick: Switched!

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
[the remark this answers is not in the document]
krishnaglick: Oh! `initialState` was what I was missing. Added, thanks!!

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
```diff
@@ -30,6 +32,7 @@ const ConnectionTable: React.FC<ConnectionTableProps> = ({ data, entity, onClick
   const navigate = useNavigate();
   const query = useQuery<{ sortBy?: string; order?: SortOrderEnum }>();
   const allowAutoDetectSchema = useFeature(FeatureItem.AllowAutoDetectSchema);
+  const streamCentricUIEnabled = useExperiment("connection.streamCentricUI.v2", false);
```
krishnaglick, line 35: Could name this `streamCentricUIv2Enabled`

airbyte-webapp/src/components/EntityTable/components/StreamStatusCell.tsx
krishnaglick: I'm mixed on what I renamed this to.

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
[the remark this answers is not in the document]
dizel852: Anytime 😉

PR #150, opened 2023-02-26
Fix resource path for secret mask

jdpgrailsdev:
## What
* Ensure that the generated secret mask file for Log4j is loaded

## How
* Use the correct resource path format to the file on the classpath by adding the missing leading "/".

## Recommended reading order
1. `AirbyteCatalogConstants.java`

- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
🚨🚨 Restores previous functionality that ensured that any field marked as secret in the Airbyte catalog is masked when logged.


cpdeethree (approved):

evantahler (approved):
Thanks @jdpgrailsdev!

davinchia (approved):

davinchia:
The latest commit is a regression test for retrieving the mask I've confirmed passes locally. Instead of waiting for the entire build to pass, I'm going to merge this in and start the process of deploying this to prod.

PR #153, opened 2023-02-27
Helm chart publish trigger fix

supertopher:
master is now main because we moved repos

this will enable builds to trigger when they hit main

### DIFF
--- a/.github/workflows/publish-helm-charts.yml
+++ b/.github/workflows/publish-helm-charts.yml
@@ -5,7 +5,7 @@ concurrency:
 on:
   push:
     branches:
-      - "master"
+      - "main"
     paths:
       - "charts/**"
       - "!charts/**/Chart.yaml"


supertopher:
merge-person: this is already broken on prod and shouldn't affect anything but pipelines

sh4sh:
Pushed the current `v0.44.1-helm` [tag to this repo](https://github.com/airbytehq/airbyte-platform/releases/tag/v0.44.1-helm) as the workflow uses it to determine next semver.

When I looked for the appropriate commit to base the tag on, I observed that chart versions in this repo are currently outdated: 
https://github.com/airbytehq/airbyte-platform/blob/9748479170cbd40c6f44db5e066715e2af4b9518/charts/airbyte/Chart.yaml#L18
The [v0.44.1-helm tag](https://github.com/airbytehq/airbyte/releases/tag/v0.44.1-helm) pre-monorepo was based off an unmerged branch: https://github.com/airbytehq/airbyte/pull/22964 I guess it doesn't have to be merged to package/release the chart, but it does mean that we're divergent. :(

Next helm publish we can't rely on chart versions in main and need to bump to a version *after* v0.44.1.

PR #154, opened 2023-02-27
change refs/head/master to refs/head/main

cpdeethree:
## What
Updates gradle.yml workflows

## How
find/replace


cpdeethree:
This is part of the run oss builds in cloud and as such is no longer necessary since it was taken care of in another pr

PR #157, opened 2023-02-27
Add Monitoring to the heartbeat timeout

benmoriceau:
## What
Add monitoring in order to see which source are timing out. This is adding 1 new datadog metric that will be send once per source which is timing out.

## Recommended reading order
1. `OssMetricsRegistry.java` => the new metric
2. `HeartBeatTimeoutChaperone.java` (and test) => send the metric to datadog
3. Other files are about properly creating the chaperone.


jdpgrailsdev (approved):
:shipit:

PR #160, opened 2023-02-27
:window: :tada: show FCP enrollment banner on create connector page

ambirdsall:
## What
Shows the FCP enrollment banner on the create connection page. 

## Awkward detour
The existing create connection page uses `location.state` to track the progressing  connector definition, whether they are selecting a source or destination, and so on; but as `location.state` is an ephemeral, in-memory thing, it gets blown away by the enrollment workflow's round-trip visit to an external stripe domain. There would be a ton of benefits to rewriting the routing and control-flow wholesale, with the url changing with the workflow steps, functioning history navigation, and the already-selected connector ids referenced in path params like `/workspaces/39f35e75-3719-44c5-9ed1-03a82144d141/sources/7b63a93f-c296-45ac-965b-6988a023ec9b/connections/new` or whatever instead of `location.state`*, I decided the path of least resistance would be to just slap the connectorIds in a query param and keep it in sync with `location.state` (the source of truth if present) with `useEffect`.

There were enough confusing threads of logic to follow and fussy interactions between `location.state` and url changes that I'm not sure how much effort I actually saved by working around the existing routing scheme.

\* and hey, as long as I'm wishing,  a reducer to advance through workflow steps from different starting points by dispatching actions instead of imperative code would be nice too.

### does it show the new banner when it ought to?
A quick legend for the emojis to follow: the TL;DR is that 🚧  means "incorrect", and check marks mean "correct".
|              | did show | no show |
|---|---|---|
| should | :heavy_check_mark: | :construction: : |
| no plz  | :construction: : | :ballot_box_with_check: |
#### :ballot_box_with_check: new workspace, creating first source, GA source
image

#### :heavy_check_mark: new workspace, creating first source, alpha/beta
image

#### :heavy_check_mark: creating second source, GA connector, first was alpha/beta
image

#### :heavy_check_mark: creating connector with prefilled destinationId
image


ambirdsall:
Where things stand, per my memory after being sick for a week:
- I have not yet addressed [the apparent race condition Teal noticed](https://www.loom.com/share/ebfc1ad1cc2249ef9ef71d8e6e136dc7), where a user's first alpha/beta source might not "register" in time to show the banner on the first render, causing a weird layout jump after clicking the dropdown (or anything else that would cause a rerender, presumably). I haven't reproduced it yet, so I don't know for sure if it's a frontend or backend fix (I'm pretty confident it's frontend, though).
- we all agree that the holdover code is gnarly and gross and that a ground-up rewrite is in order soon, but shipping this feature probably shouldn't be held up on that

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
```diff
@@ -34,16 +34,20 @@ export const useFreeConnectorProgram = () => {
   const { registerNotification } = useNotificationService();
   const { trackError } = useAppMonitoringService();
 
+  const removeStripeSuccessQuery = () => {
+    const { [STRIPE_SUCCESS_QUERY]: _, ...unrelatedSearchParams } = Object.fromEntries(searchParams);
+    setSearchParams(unrelatedSearchParams, { replace: true });
+  };
```
tealjulia, line 40: Sheer curiosity here -- is there a reason we _need_ to clean this up or is this just a good practice not keeping things we don't need and keeping URLs neat?

airbyte-webapp/src/pages/connections/CreateConnectionPage/CreateConnectionPage.tsx
```diff
@@ -251,6 +213,7 @@ export const CreateConnectionPage: React.FC = () => {
               }
             />
           )}
+          {currentStep !== StepsTypes.CREATE_CONNECTION && <InlineEnrollmentCallout withBottomMargin />}
```
tealjulia, line 216: What is blocking us from having this on the CREATE_CONNECTION step?

tealjulia (approved):
Tested it out again locally and it looks good to me.

Had a few questions but I don't think they're blocking.

airbyte-webapp/src/pages/connections/CreateConnectionPage/usePreloadData.ts
```diff
@@ -0,0 +1,113 @@
...
+  DestinationRead,
+  SourceDefinitionRead,
+  SourceRead,
+} from "core/request/AirbyteClient";
+import { useGetDestination } from "hooks/services/useDestinationHook";
+import { useGetSource } from "hooks/services/useSourceHook";
+import { useDestinationDefinition } from "services/connector/DestinationDefinitionService";
+import { useSourceDefinition } from "services/connector/SourceDefinitionService";
+
+export function hasSourceId(state: unknown): state is { sourceId: string } {
```
krishnaglick, line 15: [airbyte-webapp/src/components/ui/Text/TextWithOverflowTooltip.tsx](https://github.com/airbytehq/airbyte/pull/22701#discussion_r1107747584)

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
ambirdsall: Largely just being cautious: if we don't clean up the query string, they'd get the whole popup confirmation experience again if they were to refresh the page, which would be weird.

airbyte-webapp/src/pages/connections/CreateConnectionPage/CreateConnectionPage.tsx
ambirdsall: That's the final step in the flow before hitting the actual connection page, after creating both the source and destination, where you configure name, syncing, etc; it's the one where the page spacing is totally different because for some reason the original author decided to abandon the wrapper element every other step uses for `MainPageWithScroll`, so fixing the alignment means adding some hard-to-understand coupling to the CSS of either `CreateConnectionPage` or `InlineEnrollmentCallout`.

airbyte-webapp/src/pages/connections/CreateConnectionPage/usePreloadData.ts
ambirdsall: [TextWithOverflowTooltip.tsx](https://github.com/airbytehq/airbyte/pull/22701#discussion_r1120975697)? 😂 

For future readers, KC is linking to a question about improving the type signature for `state` in the original PR in the `airbyte` repo; in response, I'm linking to my answer, which is that "better" type signatures conflict with the actual `state` value, which is `location.state`, which is a web API with a predefined type of `unknown`, which is a type which exists specifically to refuse to match anything helpful.

airbyte-webapp/src/pages/connections/CreateConnectionPage/CreateConnectionPage.tsx
ambirdsall: Strictly speaking, this check is redundant, because of lines 189-195 above:
```ts
  if (currentStep === StepsTypes.CREATE_CONNECTION) {
    return (
      <MainPageWithScroll headTitle={headTitle} pageTitle={pageHeader}>
        {renderStep()}
      </MainPageWithScroll>
    );
  }
```
but I think it's best to leave the redundant check in to make regressions less likely in the inevitable refactoring: the question is not whether this component will be rewritten, but how soon and by whom.

ambirdsall: Hah, in fact, the redundant update is now being reported as a typescript error, so I'll remove it anyway.

krishnaglick (approved):
Diff looks wild. I think you'll need to port this to `-internal` now :/

PR #162, opened 2023-02-28
[Connector Builder Cloud] Base Micronaut application for the Connector Builder Server

brianjlai:
## What
tbd

## How
tbd
## Recommended reading order
1. `x.java`
2. `y.java`

Yes, and it should also be in a completely isolated module that is not being exposed anywhere in our system


brianjlai:
closing since this will be done on internal

PR #163, opened 2023-02-28
strict_comparison normalization: Use a correctly-configured LD flag

edgao:
## What
switch to a flag that works.

new flag - https://app.launchdarkly.com/default/production/features/normalization.strictComparisonEnabled

old flag (archived) - https://app.launchdarkly.com/default/production/features/normalization.strictComparisonNormalizationEnabled

note that the new flag has served nonzero requests recently (that was me testing locally). Cole also helped me out with writing the targeting rule correctly. (for now, we need to duplicate the user.workspace and workspace.key rules; after this week(ish) we'll only need the workspace.key rule)

also, drop a redundant `toString` call on the workspace UUID object.

## How
frankly, I wish I knew what the difference was. The settings are identical. (the targeting rules are different, of course - but that shouldn't affect the flag's ability to receive requests)

- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
none


jbfbell (approved):

jcowanpdx (approved):
LGTM

grishick (approved):

PR #164, opened 2023-02-28
Use sync persistence

gosusnp:
## What

This PR prepares us to enable persisting states from the replication rather than wait for PersistStateActivity.
* Contains the micronaut related changes to makes SyncPersistence injectable
* Adds metrics to track usage of the current state persistence from PersistStateActivity as well as the new workflow
* Updates StandardSyncInput and StandardSyncOutput with the commit state feature flag.
* New SyncPersistence should be protected by a feature flag so no expected behavior change

If the flag gets turned on:
* we should stop persisting states from the PersistStateActivity and start persisting from the DefaultReplicationWorker (runs in airbyte-workers in docker, runs in orchestrator in kube).
* For consistency to avoid having the flag switch in the middle of a sync, the flag is read at the beginning of the sync, and passed along from the SyncInput.

Closes https://github.com/airbytehq/airbyte/issues/23047

## Recommended reading order
1. `airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java` contains the new usage
2.  `airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/PersistStateActivityImpl.java` contains the flag to disable the old behavior
3. `airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/sync_persistence/SyncPersistenceImpl.java` has been updated with metrics

*If unsure, leave it blank.*
- [V] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨

Change should be protected by a feature flag, no expected behavior change other than a new metrics to track current state persist operation.


airbyte-container-orchestrator/src/main/resources/application.yml
```diff
@@ -25,6 +25,9 @@ airbyte:
         name: ${AIRBYTE_API_AUTH_HEADER_NAME:}
         value: ${AIRBYTE_API_AUTH_HEADER_VALUE:}
       host: ${INTERNAL_API_HOST}
+  worker:
+    replication:
+      persistence-flush-period: ${REPLICATION_FLUSH_PERIOD:60}
```
jdpgrailsdev, line 30: Nit:  consider adding the unit type to the config name/env var (e.g. `-sec`, `_SECONDS`).

jdpgrailsdev (commented):
LGTM.  It appears that there are a few failing tests that need to be addressed.  It would also be nice to add a bit more documentation to the code about the use of `@Prototype` for those who aren't as familiar with it.

We might also want to consider creating a custom APM trace on the `SyncPersistenceImpl` so that we will see that code appear in any existing activity traces as a separate line item.

airbyte-container-orchestrator/src/main/resources/application.yml
gosusnp: done

gosusnp:
@jdpgrailsdev, I added a few lines about the expected scope of the `SyncPersistenceImpl` and the use of `@Prototype`.

About the tracing, that makes sense, I am trying to understand what's a good tag, looking at the `operationName`, what's the current strategy for tracing?

jdpgrailsdev:
> About the tracing, that makes sense, I am trying to understand what's a good tag, looking at the `operationName`, what's the current strategy for tracing?

 @gosusnp The general strategy is to group logical operations together for items that are top level.  In this case, your code is being called by an activity, that is already the root of a trace, so it doesn't actually matter too much in this case.  That being said, if we want to keep things clean, I would probably make the operation be something like "sync" and let the tracer automatically name the resource after the class and method being traced.  The goal here is to get the invocation o the persistence to show up as a line item in the trace under the calling activity.

gosusnp:
@jdpgrailsdev, I am going to let the tracer pick the default for now.
When thinking about how to keep traces organized, I am wondering if having a top level trace that would be the `SyncWorkflow` would make sense. The current operations are a mix of different natures, now that tracing is in place, it could be worth quickly reassessing what we have.

gosusnp:
Superseeded by https://github.com/airbytehq/airbyte-platform-internal/pull/4744/

PR #165, opened 2023-02-28
Fix inputs/outputs of webapp tasks

timroes:
## What

Fix the input/outputs of all webapp gradle tasks to make them build more incremental.

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

### DIFF
--- a/airbyte-webapp/build.gradle
+++ b/airbyte-webapp/build.gradle
@@ -27,6 +27,12 @@ node {
 
 task validateLockFiles {
     description "Validate only a pnpm-lock.yaml lock file exists"
+    inputs.files "pnpm-lock.yaml", "package-lock.json", "yarn.lock"
+
+    // The validateLockFiles has no outputs, thus we always treat the outpus up to date
+    // as long as the inputs have not changed
+    outputs.upToDateWhen { true }
+
     doLast {
         assert file("pnpm-lock.yaml").exists()
         assert !file("package-lock.json").exists()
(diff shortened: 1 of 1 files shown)


airbyte-webapp/build.gradle
```diff
@@ -46,8 +52,15 @@ task pnpmBuild(type: PnpmTask) {
 
     inputs.files commonConfigs
     inputs.file '.eslintrc.js'
+    inputs.file '.stylelintrc'
+    inputs.file 'scripts/calculate-source-hash.sh'
+    inputs.file 'scripts/load-declarative-schema.sh'
```
josephkmh, line 57: Maybe worth adding the whole `scripts` folder? There are some dev scripts in there that don't matter, but I can imagine we will add some other build-relevant scripts there later and forget to add them here.

airbyte-webapp/build.gradle
```diff
@@ -46,8 +52,15 @@ task pnpmBuild(type: PnpmTask) {
...
+    inputs.file 'orval.config.ts'
+    inputs.file 'vite.config.ts'
+    inputs.file 'index.html'
```
josephkmh, line 60: Should we add `.env` to this list?

josephkmh (approved):
Looks good, just a couple suggestions of additional input files for the build

timroes:
Superseeded by https://github.com/airbytehq/airbyte-platform-internal/pull/4811

airbyte-webapp/build.gradle
timroes: `.env` is already part of that list via `commonConfigs` above.

PR #167, opened 2023-02-28
🪟 🎉 Prevent creating/saving connection when no streams are selected

edmundito:
## What
Resolves https://github.com/airbytehq/airbyte/issues/22944

Update the validation of the connection create page and the replication settings to prevent saving when no streams are selected. It shows an error to the left of the save button.

Additionally, clean up the create connection error messaging out of the old design to match the replications settings.

image

## How
* Updates the validation to test if there are any selected configured streams. Otherwise, it returns a custom message.
* Update CreateControls to remove styled-components and match the design of EditControls


airbyte-webapp/src/components/connection/ConnectionForm/formConfig.tsx
```diff
@@ -144,102 +144,96 @@ const createConnectionValidationSchema = (
...
-                id: yup
-                  .string()
-                  // This is required to get rid of id fields we are using to detect stream for edition
-                  .when("$isRequest", (isRequest: boolean, schema: yup.StringSchema) =>
-                    isRequest ? schema.strip(true) : schema
-                  ),
-                stream: yup.object(),
-                config: yup
-                  .object({
+      syncCatalog: yup.object({
```
edmundito, line 147: There was a lot of code duplication here, so the `isNewTableDesignEnabled` check is moved down to the config section, since this is where the two differ (the new table validates the pk and cursor individually)

dizel852 (approved):
LGTM 👍 
Tested locally - works as should

airbyte-webapp/src/components/connection/CreateConnectionForm/CreateConnectionForm.tsx
```diff
@@ -117,7 +117,7 @@ const CreateConnectionFormInner: React.FC<CreateConnectionPropsInner> = ({ schem
               <CreateControls
                 isSubmitting={isSubmitting}
                 isValid={isValid && !editingTransformation}
-                errorMessage={getErrorMessage(isValid, dirty)}
+                errorMessage={getErrorMessage(isValid, dirty, errors)}
```
krishnaglick, line 120: Do these errors ever have overlap with the other errors this function uses?

krishnaglick (commented):
Functionality looks good, just had one question.

airbyte-webapp/src/components/connection/ConnectionForm/CreateControls.tsx
```diff
@@ -14,49 +12,10 @@ interface CreateControlsProps {
...
-  font-size: 12px;
-  line-height: 18px;
-  color: ${({ theme }) => theme.darkPrimaryColor};
-`;
-
-const ErrorText = styled.div`
-  font-weight: normal;
-  color: ${({ theme }) => theme.dangerColor};
-  max-width: 400px;
-`;
```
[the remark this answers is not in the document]
krishnaglick: 🔥

airbyte-webapp/src/components/connection/CreateConnectionForm/CreateConnectionForm.tsx
edmundito: There is some overlap with `isValid`, but if you look at the updated implementation of this function it only checks if there is that specific stream error; otherwise it works the same way as before. I just pushed a test to verify that getErrorMessage still works as expected.

krishnaglick: Half tempted to say we should check errors for isValid, but seems like not worth doing.

letiescanciano:
@edmundito any updates on when this will be merged? I have another PR depending on this :)

edmundito:
Moved internally.

PR #176, opened 2023-02-28
Add Workspace ID tag to Connection Manager's RecordMetrics activity

pmossman:
## What
Today we tag workflow metrics with the Connection ID in the RecordMetrics activity. I want to leverage these metrics to build a Canary monitoring dashboard, where we can track and monitor particular connections that we want to classify as "canary" connections. These are connections that we know to be stable and representative of a healthy platform, so any failures from these connections should trigger a monitor that engineers can investigate.

Ideally, we can create a Canary Workspace in which every connection is actively monitored for failures. This PR will make this easy by adding the WORKSPACE_ID as an additional tag/attribute in the RecordMetrics activity, so that the Canary monitor and dashboard can query for just the Canary's workspace ID, rather than needing to specify every Connection ID in the DataDog query. 

## How
- Fetch the connection's workspaceID using the workspaceApi in the RecordMetricsImpl.
- Wrap the API call in a try/catch so that a RetryableException can be thrown in case of any API connectivity issues.
- Update activity unit tests to cover new behavior (and some refactoring to DRY it up a bit)

## Recommended reading order
Top to bottom

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
Nope


airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/MetricTags.java
```diff
@@ -14,17 +14,18 @@
...
   public static final String JOB_ID = "job_id";
   public static final String JOB_STATUS = "job_status";
   public static final String RELEASE_STAGE = "release_stage";
   public static final String RESET_WORKFLOW_FAILURE_CAUSE = "failure_cause";
-  public static final String WORKFLOW_TYPE = "workflow_type";
-  public static final String ATTEMPT_QUEUE = "attempt_queue";
-  public static final String GEOGRAPHY = "geography";
   public static final String UNKNOWN = "unknown";
+  public static final String WORKFLOW_TYPE = "workflow_type";
+  public static final String WORKSPACE_ID = "workspace_id";
```
pmossman, line 28: alphabetized these constants and added WORKSPACE_ID

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/RecordMetricActivityImpl.java
```diff
@@ -86,9 +99,23 @@ private Map<String, Object> generateTags(final ConnectionUpdaterInput connection
...
     return tags;
   }
 
+  private Optional<UUID> getWorkspaceId(final RecordMetricInput recordMetricInput) {
+    if (recordMetricInput.getConnectionUpdaterInput() == null) {
+      return Optional.empty();
+    }
+    try {
+      final UUID connectionId = recordMetricInput.getConnectionUpdaterInput().getConnectionId();
+      final WorkspaceRead workspaceRead = workspaceApi.getWorkspaceByConnectionId(new ConnectionIdRequestBody().connectionId(connectionId));
```
benmoriceau, line 114: I believe that it would worth implementing some caching here. The workspace is not expecting to change during the run so we will do many useless request. I think that a in memory cache can be possible. Another solution would be to add an activity that will fetch the workspace Id from a connection Id. We currently have ~8 calls to the record metrics which will generate 8 calls where we actually need only one.

benmoriceau (commented):
I think that the logic is ok but we can avoid some request with a cache. Maybe we could use something like: https://guava.dev/releases/17.0/api/docs/com/google/common/cache/CacheBuilder.html to implement the caching logic.

pmossman:
Ended up implementing a very basic Micronaut cache, I ran OSS Airbyte locally and added some logging to verify that we were only calling the workspaceAPI once per connection ID, and it seems to be working as expected!

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/RecordMetricActivityImpl.java
pmossman: Added @Cacheable in order to use a simple Micronaut cache for this method

davinchia: I'm not very familiar with the Cacheable annotation here. I would defer to Benoit for this.

davinchia (approved):
Good find @pmossman !

I'm not familiar with the cacheable annotation, so I'll defer to Benoit for this. Otherwise, looks good to me.

sivankumar86:
Could you let me know how to build repor-metrics pod as I am facing java cast issue and i want to fix it however, there is no document how to build the image from source. it would be helpful if you can shed some light on this ?

PR #183, opened 2023-03-01
Helm charts should be published when pushing to Main, not Master

alonsonarro:
## What
The workflow to push Helm charts used the old master branch naming scheme. However, the airbyte-platform repo is using the main naming scheme for branches. This discrepancy is causing the new Helm charts to not be published.

## How
Changing the worfklow from "on push to master" to "on push to main"

- [X ] YES 💚
- [ ] NO ❌


alonsonarro:
PR #153  is already taking care of it. Will close this one.

PR #190, opened 2023-03-02
[ISSUE #22452] create ACTIVE_DECLARATIVE_MANIFEST and DECLARATIVE_MAN…

maxi297:
…IFEST tables

## What
Addresses https://github.com/airbytehq/airbyte/issues/22452

## How
Create the tables and add tests

## Recommended reading order
1. `airbyte-config/config-models/src/main/resources/types/DeclarativeManifest.yaml` and `airbyte-config/config-models/src/main/resources/types/ActiveDeclarativeManifest.yaml` for the models
2. `airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_00_002__New_migration.java` for the tables that will be created
3.  `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java` for the operations in DB
4.  The tests


airbyte-bootloader/src/test/java/io/airbyte/bootloader/BootloaderTest.java
```diff
@@ -82,7 +82,7 @@ class BootloaderTest {
 
   // ⚠️ This line should change with every new migration to show that you meant to make a new
   // migration to the prod database
-  private static final String CURRENT_CONFIGS_MIGRATION_VERSION = "0.41.00.001";
+  private static final String CURRENT_CONFIGS_MIGRATION_VERSION = "0.41.00.002";
```
flash1293, line 85: @xiaohansong do we need to bump to `0.41.01.000` here?

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_00_002__New_migration.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.JSONB;
+import org.jooq.impl.DSL;
+import org.jooq.impl.SQLDataType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// TODO: update migration description in the class name
+public class V0_41_00_002__New_migration extends BaseJavaMigration {
```
flash1293, line 21: should be renamed to something like `AddDeclarativeManifestTables` or so

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_00_002__New_migration.java
```diff
@@ -0,0 +1,70 @@
...
+        DSL.field("created_at", SQLDataType.TIMESTAMPWITHTIMEZONE.nullable(false).defaultValue(currentOffsetDateTime()));
+    final Field<OffsetDateTime> updatedAt =
+        DSL.field("updated_at", SQLDataType.TIMESTAMPWITHTIMEZONE.nullable(false).defaultValue(currentOffsetDateTime()));
+
+    ctx.createTableIfNotExists("active_declarative_manifest")
+        .columns(id, actorDefinitionId, version, createdAt, updatedAt)
+        .constraints(
+            primaryKey(id),
+            foreignKey(actorDefinitionId, version).references("declarative_manifest", "actor_definition_id", "version"),
+            unique(actorDefinitionId))
```
flash1293, line 70: I'm just wondering now that I'm seeing it here - as actor definition id is unique, can we get rid of the synthetic `id` completely? I don't see any situation where it would help much.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -2228,6 +2233,157 @@ public void writeBuilderProject(final ConnectorBuilderProject builderProject) th
...
+   * Read all actor definition ids which have an active declarative manifest pointing to them
+   *
+   * @throws IOException exception while interacting with db
+   */
+  public Stream<UUID> getActorDefinitionIdsWithActiveDeclarativeManifest() throws IOException {
+    return database
+        .query(ctx -> ctx
+            .select(ACTOR_DEFINITION.ID)
+            .from(ACTOR_DEFINITION)
+            .join(ACTIVE_DECLARATIVE_MANIFEST, JoinType.JOIN)
```
flash1293, line 2381: we discussed to avoid joins between the new tables and the actor definition table - in this case it seems like just checking the actor definition ids from the `ACTIVE_DECLARATIVE_MANIFEST` table is sufficient, so we don't need to do this join.

maxi297: So there are no reason why we might have a ACTIVE_DECLARATIVE_MANIFEST that points to a ACTOR_DEFINITION that has been deleted, right?

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_00_002__New_migration.java
maxi297: And would you see `actorDefinitionId` as the primary key then?

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: Actor definitions should never be fully deleted (only soft-deleted via tombstone), so that should be OK

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_41_00_002__New_migration.java
flash1293: yeah, exactly. Do you see a reason against this?

maxi297: Nop, just wanted to make sure we were aligned

maxi297:
Closing since migrated using /create-oss-pr

PR #200, opened 2023-03-03
fix: add value for specifying deploy strategy type

xpuska513:
## What
* Update server chart with ability for user to specify what deployment strategy to use via values. Default to: Recreate


davinchia:
Thanks @xpuska513!

I don't think anything in Airbyte Kubernetes needs volume claims. Any idea how much effort it would take to remove those? 

If you are wondering how to check - as long as the Kubernetes acceptance tests passes, we should be good.

PR #202, opened 2023-03-06
fix: fix pod-sweeper script

xpuska513:
## What
* Fix pod-sweeper script with changing how we reference namespace
* Add env_vars, extraEnv and secrets values for pod-sweeper deployment

### DIFF
--- a/charts/airbyte-pod-sweeper/templates/deployment.yaml
+++ b/charts/airbyte-pod-sweeper/templates/deployment.yaml
@@ -44,6 +44,29 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        # Values from secret
+        {{- if .Values.secrets }}
+        {{- range $k, $v := .Values.secrets }}
+        - name: {{ $k }}
+          valueFrom:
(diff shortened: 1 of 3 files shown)


charts/airbyte-pod-sweeper/templates/deployment.yaml
```diff
@@ -44,6 +44,29 @@ spec:
...
+          valueFrom:
+            secretKeyRef:
+              name: server-secrets
+              key: {{ $k }}
+        {{- end }}
+        {{- end }}
+
+        # Values from env
+        {{- if .Values.env_vars }}
+        {{- range $k, $v := mergeOverwrite .Values.env_vars .Values.global.env_vars }}
```
supertopher, line 60: does this expansion language have `STDOUT` I feel like tracking this variables is pretty challenging, it would be nice to see what we are overwriting

xpuska513:
@mfsiega-airbyte , check the description + the issues is probably solved already, pushed latest commit.
Used Release helm object and Namespace parameter of it, so no  it should insert proper namespace value.

PR #205, opened 2023-03-14
Improve KubePortManagerSingleton logging

seanglynn-thrive:
## What
Improve logging within the `KubePortManagerSingleton` class to triage an issue where the class instance is unable to return available ports to the Worker process.

Some users are experiencing an issue with the Worker running with the container orchestrator on Airbyte OSS where the worker deployment fails to allocate ports to job pods and fails with the following exception.
```log
io.airbyte.workers.exception.WorkerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "io.airbyte.workers.process.KubePortManagerSingleton.take()"
```

As there is limited logging/tracing in this class, we would like to add more logging around this process to understand the problem at hand a little better.

Refer to the slack conversation here for context: https://airbytehq.slack.com/archives/C021JANJ6TY/p1677680873304649

## How
Add debug logs to return helpful information on how the class instance manages ports:
+ Print remaining ports when `KubePortManagerSingleton.take()` method is called.
+ Print available ports every time `KubePortManagerSingleton` instance is initialised.

## Recommended reading order
1. `KubePortManagerSingleton.java`

*If unsure, leave it blank.*
- [ ] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨


pcorbel (approved):
LGTM

justenwalker:
The reason this fails is because the `poll()` method returns `null` when the timeout is reached. 

What is happening is that there are no ports available (LinkedBlockingDeque is empty) to the worker, and it is taking more than 10 minutes (hard coded) for a port to be reclaimed.

```java
  public Integer take() throws InterruptedException {
    return workerPorts.poll(10, TimeUnit.MINUTES); // Hard-coded timeout getting next available port.
  }
```

```java
    /**
     * Retrieves and removes the head of this queue, waiting up to the
     * specified wait time if necessary for an element to become available.
     *
     * @param timeout how long to wait before giving up, in units of
     *        {@code unit}
     * @param unit a {@code TimeUnit} determining how to interpret the
     *        {@code timeout} parameter
     * @return the head of this queue, or {@code null} if the
     *         specified waiting time elapses before an element is available
     * @throws InterruptedException if interrupted while waiting
     */
    E poll(long timeout, TimeUnit unit)
        throws InterruptedException;
```

since poll returns `null` but does not throw; the implicit conversion from  `Integer` to `int` fails:

```java
      final int stdoutLocalPort = KubePortManagerSingleton.getInstance().take();
```

It would probably would be best for the worker not to take the job if it doesn't have capacity to run it ... but its not obvious (to me anyhow) how to surface that information to the worker for scheduling purposes.

seanglynn-thrive:
@justenwalker 
I've posted a new comment under the open issue here: https://github.com/airbytehq/airbyte/issues/24277#issuecomment-1497247728
It would be great if we could log the `TEMPORAL_WORKER_PORTS` that are free or log the ports are assigned to which pods (e.g. port `9001,9002` allocated to `source-read-1-0-aaaa`).

justenwalker:
It seems at least something is already logged

```
    KubePortManagerSingleton.getInstance().offer(stdoutLocalPort);
    KubePortManagerSingleton.getInstance().offer(stderrLocalPort);

    LOGGER.info(prependPodInfo("Closed all resources for pod", podDefinition.getMetadata().getNamespace(), podDefinition.getMetadata().getName()));

```
Do you see logs containing "Closed all resources for pod"?

seanglynn-thrive:
> 

Yes, I can see those logs being printed. I guess we can in fact determine what ports have been allocated to which pods and furthermore the pods that have been closed/shutdown.

Do you think it's ever possible that the `close()` method is not getting invoked for some failed/stale pods?
https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java#L731

justenwalker:
> > 
> 
> Yes, I can see those logs being printed. I guess we can in fact determine what ports have been allocated to which pods and furthermore the pods that have been closed/shutdown.
> 
> Do you think it's ever possible that the `close()` method is not getting invoked for some failed/stale pods? https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java#L731

that would explain the symptoms; but it would be difficult to track down all the places that close might *need* to be called...  

Have you tried upgrading to 0.42.0?

justenwalker:
It looks like AsyncOrchestratorPodProcess can wait up to 10 days for a pod to finish. 
This might be fine (some syncs could take a very long time).

```
  @Override
  public int waitFor() throws InterruptedException {
    final boolean exited = waitFor(10, TimeUnit.DAYS);
```

It looks like this ultimately calls `computeExitValue()` which is looking at the pod status to see if it has exited (and with what status code).

You might be right that it is not detecting `Init:Error` states correctly -- it only undestands the following:

```
  NOT_STARTED, // Pod hasn't been started yet.
  INITIALIZING, // On-start container started but not completed
  RUNNING, // Main container posted running
  FAILED, // Reported status was "failed" or pod was in Error without a reported status.
  SUCCEEDED; // Reported status was "success" so both main and on-start succeeded.
```

The default is to assume it is running; so if it was unable to determine the correct status, it would likely assume it is running and wait for the full 10 days before forcing it closed.

```
    switch (docStoreStatus) {
      case NOT_STARTED -> throw new IllegalThreadStateException("Pod hasn't started yet.");
      case INITIALIZING -> throw new IllegalThreadStateException("Pod is initializing.");
      default -> throw new IllegalThreadStateException("Pod is running.");
    }

// NOTE: it looks exceptions are used as flow control here... which is is an odd choice.
  public boolean hasExited() {
    try {
      exitValue();
      return true;
    } catch (final IllegalThreadStateException e) {
      return false;
    }
  }
```

However, if you have the pod sweeper running, the pod sweeper should see pods in an Error state and remove them after they age 24 hours i think.

davinchia:
let me handle this!

supertopher (approved):
I love it

perangel:
This is now merged. Thank you for contributing the PR!

PR #212, opened 2023-03-31
Upgrade temporal to latest release

vr:
This is the following of https://github.com/airbytehq/airbyte/pull/21978 @marcosmarxm 

Best will be to upgrade to latest temporal release, but at least bump to latest v1.13.x to get rid of the CVE.


marcosmarxm:
Thanks @vr for the update. @davinchia / @benmoriceau is it possible to take a look and execute some tests?

vr:
BTW the image is vulnerable by itself because it has been built a long time ago, i think we really need to update it on a more regular schedule.
image
image

davinchia:
@vr thanks! we ware working through some of our tooling now. I will get this merged in once that is done, probably at the start of next week.

vr:
Looking at the 1.13.4 Dockerfile this will not fix all the CVE because it depends on a version of docker.io/temporalio/base-server which is one year old.

It looks like the goal of rebuilding our own base for temporal image is useless now because they now build arm64 image so we can just use their base image, it will be way simpler for security upgrade.

I updated my commit.

I tried to test it but i can't get the test script working, can you check it maybe ?

vr:
any updates on this ?

davinchia:
This is failing internal tests. There was a fire on my street last week which took out the internet so I haven't had a chance to look at this yet. Will update you.

vr:
any updates here @davinchia ?

vrfn:
Hi @davinchia, would be great if this gets merged as the version that is currently used is ancient 😄

marcosmarxm:
@vr and @vrfn there were some attempts to upgrade temporal to latest version but it wasn't possible. 
@pmossman do you have information to share about this?

imonacloud:
any updates here? this is an issue for airbyte users deploying with helm charts + terraform where temporal v1.13.0 (which we use now) can't handle temporal and temporal_visibility dbs being pre-created but v1.15.0 of temporal can

flutra-osmani:
This change was merged; closing the ticket.

PR #220, opened 2023-04-07
fix(airbyte-config): use SELECT 1 as health check query

justenwalker:
## What

In some cases, `SELECT id FROM workspaces LIMIT 1` can a long time to return because it is a query on an active table with real data. Even if there is a limit applied, it uses a Seq Scan on the table. I've seen this take a minute and a half to return sometimes (although mostly it returns in a few milliseconds)

```
airbyte=> explain select id from workspace limit 1;
                            QUERY PLAN
------------------------------------------------------------------
 Limit  (cost=0.00..0.26 rows=1 width=16)
   ->  Seq Scan on workspace  (cost=0.00..10.40 rows=40 width=16)
```

In Kubernetes, when this endpoint times out or is delayed significantly it causes the liveness probe to fail; which in turn causes the server instance to be killed and restarted; resulting in intermittent outages.

## How

Use `SELECT 1` as the query instead.

```
airbyte=> explain select 1;
                QUERY PLAN
------------------------------------------
 Result  (cost=0.00..0.01 rows=1 width=4)
(1 row)
```

- [x] YES 💚

## 🚨 User Impact 🚨

None


cgardens:
@justenwalker i plan to merge this today. in a holding pattern because of a GCP outage in the EU, but once things are nominal, i'll merge it.

justenwalker:
> @justenwalker i plan to merge this today. in a holding pattern because of a GCP outage in the EU, but once things are nominal, i'll merge it.

Hope the outage wasn't too bad. Did you still want to merge this as-is? Or want me to rebase.

justenwalker:
I see this has been merged in https://github.com/airbytehq/airbyte-platform/commit/c9a0a6f453f61a14fa9f6db22230219a833747c7 so im closing it.

PR #231, opened 2023-04-20
feat(helm): allow job images to be overwritten

justenwalker:
## What

The worker process supports overriding the images used by the kubernetes jobs; however
the helm chart so far has not had the direct ability to override values, so the defaults are always used.

This is a problem since docker hub rate limits access to public images, and it can cause an outage if those rate limits are tripped.  

```
Warning  Failed     10m (x4 over 12m)     kubelet            Failed to pull image "busybox:1.28": rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/library/busybox:1.28": failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/busybox/manifests/sha256:141c253bc4c3fd0a201d32dc1f493bcf3fff003b6df416dea4f41046e0f37d47: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
 ```

The solution is to have a local mirror or to use another public source that doesn't have rate limits.

## How

Creates a new set of values for overriding the images that are used by the worker jobs:
  - global.jobs.kube.images.busybox
  - global.jobs.kube.images.socat
  - global.jobs.kube.images.curl

*Note* : This PR is not strictly necessary to solve the issue. You can pass the values as extra environment variables; its just less discoverable without these options; implying that this as a niche concern. In reality, the docker hub rate limits are such that anyone using Airbyte at some non trivial scale will run into this issue and likely should mirror these images in their local registries and use them instead of the public ones.

```
worker:
  extraEnv:
    - name: JOB_KUBE_SOCAT_IMAGE
      value: xxxxxxxxxxxxxx.dkr.ecr.us-west-2.amazonaws.com/alpine/socat:1.7.4.3-r0
    - name: JOB_KUBE_BUSYBOX_IMAGE
      value: xxxxxxxxxxxxxx.dkr.ecr.us-west-2.amazonaws.com/library/busybox:1.28
    - name: JOB_KUBE_CURL_IMAGE
      value: xxxxxxxxxxxxxx.dkr.ecr.us-west-2.amazonaws.com/curlimages/curl:7.83.1
    - name: AWS_ACCESS_KEY_ID
```

## Recommended reading order
1. `charts/airbyte-worker/values.yaml`
2. `charts/airbyte-worker/templates/deployment.yaml`
3. `charts/airbyte/values.yaml.test`
4. `charts/airbyte/templates/env-configmap.yaml`

- [ ] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
As this is a new feature, there should be no backwards incompatibility

### DIFF
--- a/charts/airbyte-worker/README.md
+++ b/charts/airbyte-worker/README.md
@@ -1,6 +1,6 @@
 # worker
 
-![Version: 0.39.36](https://img.shields.io/badge/Version-0.39.36-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.40.4](https://img.shields.io/badge/AppVersion-0.40.4-informational?style=flat-square)
+![Version: 0.44.7](https://img.shields.io/badge/Version-0.44.7-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.44.2](https://img.shields.io/badge/AppVersion-0.44.2-informational?style=flat-square)
 
 Helm chart to deploy airbyte-worker
 
@@ -21,6 +21,8 @@ Helm chart to deploy airbyte-worker
 | extraContainers | list | `[]` |  |
 | extraEnv | list | `[]` |  |
 | extraInitContainers | list | `[]` |  |
+| extraLabels | object | `{}` |  |
+| extraSelectorLabels | object | `{}` |  |
 | extraVolumeMounts | list | `[]` |  |
 | extraVolumes | list | `[]` |  |
 | global.credVolumeOverride | string | `""` |  |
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by justenwalker** [comment]
I tested this chart out in my dev environment and it was able to pass the environment variables through to the worker. Workers then began spawning jobs with the appropriate images.

**Comment by justenwalker** [comment]
This PR is not strictly necessary to solve the issue: you can pass the values as extra environment variables; its just less discoverable without these options.

```
worker:
  extraEnv:
    - name: JOB_KUBE_SOCAT_IMAGE
      value: public.ecr.aws/tvlk/alpine/socat:1.7.3.4-r0
    - name: JOB_KUBE_BUSYBOX_IMAGE
      value: public.ecr.aws/docker/library/busybox:1.35
    - name: JOB_KUBE_CURL_IMAGE
      value: public.ecr.aws/metaplay/curlimages/curl:7.85.0
```

File: charts/airbyte-worker/README.md [hunk 1]
```diff
@@ -21,6 +21,8 @@ Helm chart to deploy airbyte-worker
 | extraContainers | list | `[]` |  |
 | extraEnv | list | `[]` |  |
 | extraInitContainers | list | `[]` |  |
+| extraLabels | object | `{}` |  |
```
cpdeethree, line 24: Thanks for helping to keep the docs up to date!

cpdeethree (approved):
Thank you for your contribution!

justenwalker:
> /create-oss-pr

justenwalker:
> Thank you for your contribution!

I've added another commit to regenerate the docs

cpdeethree:
This is merged and is now available in https://github.com/airbytehq/airbyte-platform/commit/f87795705315f44cf2e5c732439938e75dfbf754

PR #242, opened 2023-05-23
Allow to specify init job timeout

AKuzyashin:
## What
Current timeout of 5 minutes is to low for aws spot instances 
## How
Allow change timeout 

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨


Barsoomx (approved):
lgtm! very useful tweak for k8s

Barsoomx:
@cpdeethree can you please take a look?

cpdeethree:
Hello!

Thank you for your contribution!

We think allowing this to be configurable makes a ton of sense! We are not currently thinking it is a good idea to set the default to 5 minutes, however. We think it does absolutely make sense in your case though, since spot instances can take a long time to spin up. Could we revert that specific change back to 1 minute? After doing that, we are happy to accept your change!

cpdeethree (approved):

PR #248, opened 2023-06-07
Use extraVolumes and extraVolumeMounts values in airbyte-cron chart d…

evrudenko:
# The airbyte-cron chart accepts extraVolumes and extraVolumeMounts but doesn't use them.

## What

The airbyte-cron chart accepts the TEMPORAL_HISTORY_RETENTION_IN_DAYS environment variable and removes logs older than the value of the variable. Logs are written in other airbyte containers and if we don't provide a shared volume for airbyte-cron, then airbyte-cron won't be able to delete logs.

## How

The airbyte-cron chart will add volumeMount and volume sections if the variables are defined, so we can add a common volume and store logs in one place

## Recommended reading order
1. charts/airbyte-cron/templates/deployment.yaml

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌


malikdiarra:
Hi @evrudenko, thanks for your contribution.

I've merged your [PR](https://github.com/airbytehq/airbyte-platform/commit/6e17b8e883107d5b2101491e9c81da89cdc8d08b)

malikdiarra:
Update: I had to revert it as we are having some issue with the helm chart tests. I'm gonna try to reapply in the next couple of days.

PR #259, opened 2023-07-11
Add support for IAM Roles with AWS Secrets Manager

msaffitz:
## What

Relaxes the requirement to provide an awsAccessKey and awsSecretAccessKey when using AWSSecretManagerPersistence.  When these values are not provided the AWS SDK will fallback to the default credential provider (e.g. IAM Role).

## How
Adopts a similar approach to that taken in [airbyte-config/config-models/src/main/java/io/airbyte/config/storage/DefaultS3ClientFactory.java](https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-config/config-models/src/main/java/io/airbyte/config/storage/DefaultS3ClientFactory.java) as part of https://github.com/airbytehq/airbyte-platform/commit/37ba07b5001f825abaa2191422be7b04f72fff4d#

- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
The only anticipated user impact is in the event that no valid credentials are discoverable by the AWS SDK.  This would also have previously led to failure, but earlier and more explicitly in the service boot sequence.

### DIFF
--- a/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/split_secrets/AWSSecretManagerPersistence.java
+++ b/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/split_secrets/AWSSecretManagerPersistence.java
@@ -68,24 +68,23 @@ public class AWSSecretManagerPersistence implements SecretPersistence {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


pmossman:
@msaffitz thank you for the contribution! I verified that your PR passes our internal builds and merged it in. I'll close this PR as the internal merge will be mirrored back to the main branch soon.

PR #260, opened 2023-07-13
Helm - Allow customization of WEBAPP_URL

AndreaGiardini:
## What
This change allows setting the `WEBAPP_URL` variable from the helm chart. It solves the issue discussed in https://github.com/airbytehq/airbyte/issues/23269

Let me know what you think about it

### DIFF
--- a/charts/airbyte/templates/env-configmap.yaml
+++ b/charts/airbyte/templates/env-configmap.yaml
@@ -61,7 +61,7 @@ data:
   TEMPORAL_HOST: {{ .Release.Name }}-temporal:{{ .Values.temporal.service.port }}
   TEMPORAL_WORKER_PORTS: 9001,9002,9003,9004,9005,9006,9007,9008,9009,9010,9011,9012,9013,9014,9015,9016,9017,9018,9019,9020,9021,9022,9023,9024,9025,9026,9027,9028,9029,9030,9031,9032,9033,9034,9035,9036,9037,9038,9039,9040
   TRACKING_STRATEGY: segment
-  WEBAPP_URL: http://{{ .Release.Name }}-airbyte-webapp-svc:{{ .Values.webapp.service.port }}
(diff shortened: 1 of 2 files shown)


charts/airbyte/templates/env-configmap.yaml
```diff
@@ -61,7 +61,7 @@ data:
   TEMPORAL_HOST: {{ .Release.Name }}-temporal:{{ .Values.temporal.service.port }}
   TEMPORAL_WORKER_PORTS: 9001,9002,9003,9004,9005,9006,9007,9008,9009,9010,9011,9012,9013,9014,9015,9016,9017,9018,9019,9020,9021,9022,9023,9024,9025,9026,9027,9028,9029,9030,9031,9032,9033,9034,9035,9036,9037,9038,9039,9040
   TRACKING_STRATEGY: segment
-  WEBAPP_URL: http://{{ .Release.Name }}-airbyte-webapp-svc:{{ .Values.webapp.service.port }}
+  WEBAPP_URL: {{ if .Values.webapp.url }}{{ .Values.webapp.url }}{{ else }}http://{{ .Release.Name }}-airbyte-webapp-svc:{{ .Values.webapp.service.port }}{{ end }}
```
pmossman, line 64: @AndreaGiardini can you change this to use [default](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) instead of an inline if-else? That'll be more conventional I think

pmossman (changes requested):

charts/airbyte/templates/env-configmap.yaml
AndreaGiardini: @pmossman It should be fixed now. Can you have a second look, please?

pmossman: Thanks for making that change, I will open an internal pull request where I can validate the change further

pmossman:
@AndreaGiardini I merged this PR into our internal repository, so the change will be mirrored back to our public repos soon. This will be available in the next helm chart version. Thanks for the contribution!

PR #268, opened 2023-08-17
Fix/update api port

Santhin:
## What
It repairs `airbyte-api-server` port and adds missing ingress

## How
*Describe the solution*

*If unsure, leave it blank.*
- [ ] YES 💚
- [X] NO ❌

## 🚨 User Impact 🚨
No


Santhin:
Actually, you can but you need to create additional ingres by your hand, for PoC port forward can works as well

PR #269, opened 2023-08-17
fix(helm): s3PathStyleAccess should be true when minio endpoint is set

ogirardot:
## What
The change is about the s3PathStyleAccess variable that disregards the endpoint definition and should take it into account just like the minio configuration just above to set the style path to a s3 style in case of using minio endpoints.

## How
Simply add to the ternary operation the endpoint variable

- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
No breaking changes and just a bug fix


colesnodgrass:
Closing as this functionality is now determined entirely by the [`global.storage.type` value being set to `minio`](https://github.com/airbytehq/airbyte-platform/blob/3ebf8293cc17dfd7e1930a0dc0f94f50dc04cc81/charts/airbyte/templates/_helpers.tpl#L214-L216) instead of being based off of any other property.

PR #270, opened 2023-08-18
Helm: Add ability to set resources for postgresql and minio deployments

milesarmstrong:
## What
Adds the ability for users of the [airbyte Helm chart](https://docs.airbyte.com/deploying-airbyte/on-kubernetes-via-helm) to set resources for the PostgreSQL and Minio deployments.

## How
Helm templating / values.

Doesn't set any default values, so as not to change the existing behaviour of having no resources set.

## Recommended reading order
N/A

- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨


milesarmstrong:
However, a manual templating of the chart using `./ci.sh template` looks correct to me.

PR #272, opened 2023-09-06
Fix extraContainers in Cron helm chart

nightmareartist:
## What
If current chart is used with `extraContainers` enabled the deployment will fail with the following error message:

```
Error: YAML parse error on cron/templates/deployment.yaml: error converting YAML to JSON: yaml: line 94: did not find expected key
```

This is not a problem when global `extraContainers` is used, but if, for example, we want Cron specific sidecar containers the setup will fail.

## How
Added proper indentation to line 154.

*If unsure, leave it blank.*
- [X ] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
None.

### DIFF
--- a/charts/airbyte-cron/templates/deployment.yaml
+++ b/charts/airbyte-cron/templates/deployment.yaml
@@ -151,7 +151,7 @@ spec:
           {{- end }}
 
         {{- if .Values.extraContainers }}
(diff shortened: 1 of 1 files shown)


PurseChicken:
This is still an active issue. Could we please get this merged in and fixed? The proposed change from nindent 6 to nindent 8 is valid.

Thanks!

PR #276, opened 2023-09-20
fix: update ingress template value for < 1.19-0

wjmolina:
## What
The original service name for `semverCompare "<1.19-0"`, which is
```
{{ $fullName }}-svc
```
is not the correct name of the service. Instead, it should be
```
{{ $.Release.Name }}-airbyte-webapp-svc
```
This is the current value for `semverCompare ">=1.19-0"`.

Strangely, the `if` part was updated in a previous PR, but the same change was not made in the `else` part (see the image below). This PR rectifies this discrepancy.

image
## How
The `serviceName` (for `semverCompare "<1.19-0"`) was changed from `{{ $fullName }}-svc` to `{{ $.Release.Name }}-airbyte-webapp-svc`.
- [X] YES 💚
- [ ] NO ❌
## 🚨 User Impact 🚨
There are no breaking changes. The user (of this version) currently cannot access the UI. After this is merged, the user will be able to access the UI again.


marcosmarxm:
@cpdeethree can you chec this contribution?

cpdeethree:
@marcosmarxm i think this got lost in the mix, if you want to route it back through the process

wjmolina:
> @marcosmarxm i think this got lost in the mix, if you want to route it back through the process

Thanks for bumping this!

PR #279, opened 2023-10-09
Add AWS SDK's sts dependency to relevant modules to fix IRSA

sidharrth2002:
## What
Add `software.amazon.awssdk:sts` dependency to potentially resolve error caused by AWS `v2` SDK.

The relevant error [here](https://github.com/airbytehq/airbyte-platform/pull/274#issuecomment-1738864018) was raised by @larrywax, relating to using IRSA credentials. Namely, this line was in the error log:

```
WebIdentityTokenCredentialsProvider(): To use web identity tokens, the 'sts' service module must be on the class path.,
```

## How
Airbyte's use of the AWS `v2` SDK for S3 data retrieval might require adding `software.amazon.awssdk:sts` due to a hardcoded class path in AWS's [WebIdentityTokenCredentialsProvider](https://github.com/aws/aws-sdk-java-v2/blob/3fbf13cbed80c3e8ce355963be8f7bcffc2a49ab/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtils.java#L43-L55) causing errors.

## Recommended reading order
1. `deps.toml`
2. `airbyte-commons-worker/build.gradle`
3. `airbyte-config/config-models/build.gradle`
4. `airbyte-server/build.gradle`
5. `airbyte-workers/build.gradle`

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨


airbyte-oauth/build.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
     implementation libs.appender.log4j2
     implementation libs.aws.java.sdk.s3
     implementation libs.aws.java.sdk.sts
+    implementation libs.sts
```
weichunnn, line 15: I believe we don't need this since the v2 S3 module is not used in the package

weichunnn (commented):
Thanks for bringing up the changes to resolve the missing STS module!

airbyte-oauth/build.gradle
sidharrth2002: Good catch, removing!

weichunnn:
Found other modules which are relying on `implementation libs.s3` [here](https://github.com/search?q=repo%3Aairbytehq%2Fairbyte-platform%20implementation%20libs.s3&type=code). I think we can include the STS module there as well to have better coverage. What do you think?

sidharrth2002:
Agreed @weichunnn, I've added this as well to `airbyte-commons-server` and `airbyte-container-orchestrator`

pmossman:
Thanks for the continued contributions in this area @weichunnn and @sidharrth2002! I'll open this PR in our internal repo and run tests there.

pmossman:
@sidharrth2002 @weichunnn sorry for the delay here - the internal build was failing and I didn't have time to investigate until today. Turns out there was a small issue with our PR-cloning tool and the fix was simple, so I went ahead and merged the internal PR. The changes will be reflected back to this public repo, and the fix will be available in the next OSS release. Thanks again!

PR #281, opened 2023-10-20
fix: move external default docker images to latest stable

vr:
## What
Upgrade external docker image to latest stable release
better to have sane/secure defaults

- [X] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
No


colesnodgrass:
We periodically update these dependencies to use more recent versions. However we cannot use `latest` as it adds unpredictability and goes against our desire for reproducible builds.

PR #285, opened 2023-11-02
Fix inaccessible GCS bucket storage

anaselmhamdi:
## What

GCS storage is unusable when using chart version `0.49.6`.
Issue is described here as well: https://github.com/airbytehq/airbyte/issues/25906

## How

It seems that the worker would still pickup the default s3 "airbyte-dev-logs" bucket when available.
Setting it to an empty string fixed the issue.

## Recommended reading order
1. `values.yml`

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨

### DIFF
--- a/charts/airbyte/values.yaml
+++ b/charts/airbyte/values.yaml
@@ -70,7 +70,7 @@ global:
     ##  logs.s3.bucketRegion Region of the bucket (must be empty if using minio)
     s3:
       enabled: false
-      bucket: airbyte-dev-logs
+      bucket: ""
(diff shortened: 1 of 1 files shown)


colesnodgrass:
Closing as the entire external-storage configuration and code was [recently rewritten](https://github.com/airbytehq/airbyte-platform/commit/b2a5ab6cc7836ed6c1cd4cb502dcd37f384723a6).  This PR should no longer be necessary.

PR #288, opened 2023-11-22
remove replica count on worker when hpa enabled

anaselmhamdi:
## What

The current `hpa` parameter in the `values.yml` doesn't actually do anything for the worker. Worse, when you try to apply your own HPA configuration, the `replicaCount` blocks the scaling altogether.

I'm not proposing extra config for the hpa, simply the possibility to remove it if you use your own

Relates to this: https://github.com/airbytehq/airbyte/issues/25141

## How

Added an if/else removing `spec.replicas` on the worker deployment if `worker.hpa.enabled` is true 

## Recommended reading order
1. `x.java`
2. `y.java`

*If unsure, leave it blank.*
- [ ] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨

### DIFF
--- a/charts/airbyte-worker/templates/deployment.yaml
+++ b/charts/airbyte-worker/templates/deployment.yaml
@@ -8,7 +8,9 @@ metadata:
     {{ toYaml (mergeOverwrite .Values.extraLabels .Values.global.extraLabels) | nindent 4 }}
     {{- end }}
 spec:
+  {{- if eq .Values.hpa.enabled false}}
   replicas: {{ .Values.replicaCount }}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "airbyte.selectorLabels" . | nindent 6 }}


stevenmurphy12:
Hi @anaselmhamdi , I'm curious if you took this any further after closing your PR. 

I'm just starting to look into performance tuning for my Airbyte deployment. It appears that with the lack of HPA capability, my only option is to over-provision the workers for now.

alexlopes:
Same, HPA is a must-have option here. @stevenmurphy12, did you over-provision the capacity since then?

stevenmurphy12:
@alexlopes - Yes over-provisioned, though our usage level is still fairly light at the moment. By over-provisioned, I mean just two workers.

My understanding is that an Airbyte worker can handle 5 sync jobs concurrently, though that can be tuned with an environment variable. Our jobs run at different times anyway so I've not done a proper load test of what would happen if 10+ jobs happened to kick off at the same time, presumably they'd get queued in Temporal. A delay is probably acceptable tbh.

stevenmurphy12:
@alexlopes I played with `MAX_SYNC_WORKERS`, setting it to 1 had the expected effect of running 1 sync at a time and queueing the rest.

For the time being I've moved away from focusing on the worker auto-scaling, instead focussing on the resource requirements of the ephemeral pods being spun up as part of a sync. I plan to run these in their own nodepool with an appropriate node size for their K8s requests/limits. 

Something to consider with your

> heavy ETL processes

Is whether by heavy you mean a lot of syncs running concurrently, the resource requirements of the syncs being heavy, or both. The source and destination connector definitions have the option of setting K8s resource requests/limits on them (through the config API, not exposed on the UI), though I've not got as far as experimenting with those yet.

PR #289, opened 2023-11-23
Make NGNIX unprivileged docker images as default

nexcalmart:
## What
To support non-root deployments on k8s NGINX has to be configured to work without root privileges. 
There is already an official docker image of NGINX in an unprivileged mode.
Since the webapp doesn't need privileged access rights this has to be taken as the default docker image for the webapp.

## How
Change the build configuration to switch to the non-privileged docker image if not explicitly overwritten by the env variable.

## Recommended reading order
1-

- [ X ] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
No


colesnodgrass:
@perangel This PR should be closed out by your work.

PR #290, opened 2023-11-25
update minio helm template to maintain logs when it restarted

jx2lee:
## What
resolve [#27750](https://github.com/airbytehq/airbyte/issues/27750)

## How
if u want to maintain minio logs when restarted, set minio.storage.persistence `true`

## Recommended reading order
`Null`

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨


colesnodgrass:
Rather than add the ability to set a `persistence` flag, I updated the `mountPath` to match the path we actually write to (`storage`).  The `/data` path was a bug.  This also fixes an issue (one of many) with deploying to OpenShift.

jx2lee:
Great, i checked update and will close this PR

PR #293, opened 2023-12-07
fix(chart): use http scheme for INTERNAL_API_HOST

justenwalker:
## What

The current chart doesn't use a scheme for INTERNAL_API_HOST causing the api-server fail to reach the internal config API.

## How

This fix adds an explicit http:// prefix to the value of INTERNAL_API_HOST in the configmap.

## Recommended reading order
1. `charts/airbyte/templates/env-configmap.yaml`

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨

None

### DIFF
--- a/charts/airbyte/templates/env-configmap.yaml
+++ b/charts/airbyte/templates/env-configmap.yaml
@@ -24,7 +24,7 @@ data:
   DB_DOCKER_MOUNT: airbyte_db
   GCS_LOG_BUCKET: {{ .Values.global.logs.gcs.bucket | quote }}
   GOOGLE_APPLICATION_CREDENTIALS: {{ include "airbyte.gcpLogCredentialsPath" . | quote }}
-  INTERNAL_API_HOST: {{ .Release.Name }}-airbyte-server-svc:{{ .Values.server.service.port }}
+  INTERNAL_API_HOST: http://{{ .Release.Name }}-airbyte-server-svc:{{ .Values.server.service.port }}
 {{- if eq (index .Values "workload-api-server" "enabled") true }}
   # Temporary conditional for OSS deploys.  Eventually, the workload-api will be present in OSS deploys and the else
   # block can be removed


justenwalker:
Ok, so this doesn't actually work as-is. The reason is the worker expects this value in the form `host:port` so when the `http://` prefix is added, that client setup breaks: 

```
2023-12-07 20:26:16 ERROR i.m.r.Micronaut(handleStartupException):338 - Error starting Micronaut server: Error instantiating bean of type  [io.airbyte.api.client.AirbyteApiClient]

Message: For input string: "//airbyte-airbyte-server-svc"
Path Taken: new ApplicationInitializer() --> ApplicationInitializer.checkConnectionActivities --> List.checkConnectionActivities([CheckConnectionActivity checkConnectionActivity]) --> new CheckConnectionActivityImpl(WorkerConfigsProvider workerConfigsProvider,ProcessFactory processFactory,SecretsR
io.micronaut.context.exceptions.BeanInstantiationException: Error instantiating bean of type  [io.airbyte.api.client.AirbyteApiClient]

Message: For input string: "//airbyte-airbyte-server-svc"
Path Taken: new ApplicationInitializer() --> ApplicationInitializer.checkConnectionActivities --> List.checkConnectionActivities([CheckConnectionActivity checkConnectionActivity]) --> new CheckConnectionActivityImpl(WorkerConfigsProvider workerConfigsProvider,ProcessFactory processFactory,SecretsR
    at io.micronaut.context.DefaultBeanContext.resolveByBeanFactory(DefaultBeanContext.java:2367) ~[micronaut-inject-3.10.1.jar:3.10.1]
                                                                             
```

terencecho:
this issue should be fixed by this commit: https://github.com/airbytehq/airbyte-platform/commit/b28dc9c79da170e99c3d1349fbedb8117df27a82

PR #298, opened 2024-01-01
feat(api): Idempotent Resource Creation

justenwalker:
## What

Based on [[API] Idempotent Resource Creation](https://github.com/airbytehq/airbyte/discussions/33092) Discussion.

This change adds an `idempotencyKey` to the API and database tables. This enables an idempotent way to create resources; preventing duplicate creation of the same resource for retries and concurrent create requests.

## Why

And Idempotency key of some kind is necessary in order to prevent duplicates from being created in concurrent or retried requests.

For example, I often run into this scenario; especially if the Airbyte Server is under load (when it starts returning 504/502 errors):

 - Try to create a source/connection/destination/etc...
 - Get a 504/502 response back; indicating a timeout waiting for a server response.
 - Retry after some time
 - Initial request succeeds and creates a source/connection/destination/etc.
 - Retried request also succeeds and creates a source/connection/destination/etc.... with the same name

The more retries, the more duplicates.

Since no resource has any unique constraint besides its UUID (which is generated on creation) -- the server will gladly accept multiple create requests for an identical resource.

## How

1. Add a nullable uuid column called `idempotency_key` to the database tables of `organizations`, `users`, `workspaces`, `connections`, `actors`, and `actor_definitions`.
2. Add a unique index to prevent duplicate `idempotency_key` rows and fast lookup by `idempotency_key`.
3. Modify the API to accept an `idempotencyKey` value on create for these resources.
4. Create requests look up the value in the DB if it was provided, and will return the existing value instead of accepting the write.

### Database Migrations

1. airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_50_33_016__AddIdempotencyKeys.java
2. airbyte-db/db-lib/src/main/resources/configs_database/schema_dump.txt
3. airbyte-bootloader/src/test/java/io/airbyte/bootloader/BootloaderTest.java

### API Changes

1. airbyte-api/src/main/openapi/config.yaml
2. airbyte-api/src/main/openapi/api.yaml

### API Server Mappings

1. airbyte-api-server/src/main/kotlin/io/airbyte/api/server/mappers/ConnectionCreateMapper.kt
2. airbyte-api-server/src/main/kotlin/io/airbyte/api/server/services/DestinationService.kt
3. airbyte-api-server/src/main/kotlin/io/airbyte/api/server/services/SourceService.kt
4. airbyte-api-server/src/main/kotlin/io/airbyte/api/server/services/WorkspaceService.kt

### Server Handlers

1. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectionsHandler.java
2. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/DestinationDefinitionsHandler.java
3. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/DestinationHandler.java
4. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OrganizationsHandler.java
5. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandler.java
6. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceHandler.java
7. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/UserHandler.java
8. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandler.java
9. airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WorkspacesHandler.java
10. airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/DestinationDefinitionsHandlerTest.java
11. airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandlerTest.java
12. airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandlerTest.java

### Config Model Updates
1. airbyte-config/config-models/src/main/resources/types/DestinationConnection.yaml
2. airbyte-config/config-models/src/main/resources/types/IdempotencyKey.yaml
3. airbyte-config/config-models/src/main/resources/types/Organization.yaml
4. airbyte-config/config-models/src/main/resources/types/SourceConnection.yaml
5. airbyte-config/config-models/src/main/resources/types/StandardDestinationDefinition.yaml
6. airbyte-config/config-models/src/main/resources/types/StandardSourceDefinition.yaml
7. airbyte-config/config-models/src/main/resources/types/StandardSync.yaml
8. airbyte-config/config-models/src/main/resources/types/StandardWorkspace.yaml
9. airbyte-config/config-models/src/main/resources/types/User.yaml

### Persistence Changes

1. airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/OrganizationPersistence.java
2. airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/UserPersistence.java
3. airbyte-data/src/main/java/io/airbyte/data/services/ConnectionService.java
4. airbyte-data/src/main/java/io/airbyte/data/services/DestinationService.java
5. airbyte-data/src/main/java/io/airbyte/data/services/SourceService.java
6. airbyte-data/src/main/java/io/airbyte/data/services/WorkspaceService.java
7. airbyte-data/src/main/java/io/airbyte/data/services/impls/jooq/ConnectionServiceJooqImpl.java
8. airbyte-data/src/main/java/io/airbyte/data/services/impls/jooq/DestinationServiceJooqImpl.java
9. airbyte-data/src/main/java/io/airbyte/data/services/impls/jooq/OrganizationServiceJooqImpl.java
10. airbyte-data/src/main/java/io/airbyte/data/services/impls/jooq/SourceServiceJooqImpl.java
11. airbyte-data/src/main/java/io/airbyte/data/services/impls/jooq/WorkspaceServiceJooqImpl.java

- [ ] YES 💚
- [x] NO ❌

Contains API changes and DB migrations.

## 🚨 User Impact 🚨

This shouldn't be a breaking change, as the new idempotencyKey is an optional parameter; and behavior should
not be different than it currently is if it is not given in the API Calls.

### DIFF
--- a/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/DestinationDefinitionsHandler.java
+++ b/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/DestinationDefinitionsHandler.java
@@ -37,6 +37,7 @@ import io.airbyte.config.init.SupportStateUpdater;
 import io.airbyte.config.persistence.ConfigNotFoundException;
 import io.airbyte.config.persistence.ConfigRepository;
 import io.airbyte.config.specs.RemoteDefinitionsProvider;
+import io.airbyte.data.services.DestinationService;
 import io.airbyte.featureflag.DestinationDefinition;
 import io.airbyte.featureflag.FeatureFlagClient;
 import io.airbyte.featureflag.HideActorDefinitionFromList;
@@ -54,6 +55,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -68,6 +70,7 @@ import java.util.stream.Stream;
 public class DestinationDefinitionsHandler {
 
   private final ConfigRepository configRepository;
+  private final DestinationService destinationService;
   private final Supplier<UUID> uuidSupplier;
   private final ActorDefinitionHandlerHelper actorDefinitionHandlerHelper;
   private final RemoteDefinitionsProvider remoteDefinitionsProvider;
@@ -82,7 +85,8 @@ public class DestinationDefinitionsHandler {
                                        final RemoteDefinitionsProvider remoteDefinitionsProvider,
                                        final DestinationHandler destinationHandler,
                                        final SupportStateUpdater supportStateUpdater,
-                                       final FeatureFlagClient featureFlagClient) {
+                                       final FeatureFlagClient featureFlagClient,
+                                       final DestinationService destinationService) {
     this.configRepository = configRepository;
     this.uuidSupplier = uuidSupplier;
     this.actorDefinitionHandlerHelper = actorDefinitionHandlerHelper;
@@ -90,6 +94,7 @@ public class DestinationDefinitionsHandler {
     this.destinationHandler = destinationHandler;
     this.supportStateUpdater = supportStateUpdater;
     this.featureFlagClient = featureFlagClient;
+    this.destinationService = destinationService;
   }
 
   @VisibleForTesting
@@ -233,6 +238,20 @@ public class DestinationDefinitionsHandler {
 
   public DestinationDefinitionRead createCustomDestinationDefinition(final CustomDestinationDefinitionCreate customDestinationDefinitionCreate)
       throws IOException {
+    final UUID idempotencyKey = customDestinationDefinitionCreate.getIdempotencyKey();
+    if (idempotencyKey != null) {
+      final Optional<StandardDestinationDefinition> found = destinationService.getStandardDestinationDefinitionByIdempotencyKey(idempotencyKey);
+      if (found.isPresent()) {
+        final StandardDestinationDefinition destinationDefinition = found.get();
+        final ActorDefinitionVersion actorDefinitionVersion;
+        try {
+          actorDefinitionVersion = configRepository.getActorDefinitionVersion(destinationDefinition.getDefaultVersionId());
+          return buildDestinationDefinitionRead(destinationDefinition, actorDefinitionVersion);
+        } catch (ConfigNotFoundException e) {
+          // not found, so must create
+        }
+      }
+    }
     final UUID id = uuidSupplier.get();
     final DestinationDefinitionCreate destinationDefCreate = customDestinationDefinitionCreate.getDestinationDefinition();
     final ActorDefinitionVersion actorDefinitionVersion =
--- a/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandler.java
+++ b/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandler.java
@@ -38,6 +38,7 @@ import io.airbyte.config.init.SupportStateUpdater;
 import io.airbyte.config.persistence.ConfigNotFoundException;
 import io.airbyte.config.persistence.ConfigRepository;
 import io.airbyte.config.specs.RemoteDefinitionsProvider;
+import io.airbyte.data.services.SourceService;
 import io.airbyte.featureflag.FeatureFlagClient;
 import io.airbyte.featureflag.HideActorDefinitionFromList;
 import io.airbyte.featureflag.Multi;
@@ -55,6 +56,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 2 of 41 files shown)


JonsSpaghetti:
Hi @justenwalker, thank you for your contribution! We like what this change allows and discussed internally - our one question is: was there a reason a new column was created instead of using the already existing id column? That serves as our PK in our database currently and it makes sense to add that to the data models and use it as our idempotency key for this feature.

justenwalker:
@JonsSpaghetti 
> Hi @justenwalker, thank you for your contribution! We like what this change allows and discussed internally - our one question is: was there a reason a new column was created instead of using the already existing id column? That serves as our PK in our database currently and it makes sense to add that to the data models and use it as our idempotency key for this feature.

If you check the [discussion](https://github.com/airbytehq/airbyte/discussions/33092); that was an option that I presented; however my concerns are:
1. Allowing the user to supply a PK might have security implications that I cannot foresee.
2. A separate key allows potentially using this column to make some update operations idempotent too, since the ID cannot change; but the idempotency key is free to change.
3. The existence of https://airbyte-public-api-docs.s3.us-east-2.amazonaws.com/rapidoc-api-docs.html#post-/v1/workspaces/create_if_not_exist seems to imply that a user-supplied ID is only for testing, and is not a production-intended use-case
   >  Creates a workspace with an explicit workspace ID. **This should be use in acceptance tests only.**
5. The idempotency key is more ergonomic for the user ; as it allows the response to be the same for both collision and no-collision, making the client's job a lot easier; supplying the PK means that ID conflict error case must be handled separately.
6. There is some nuance around the `actors` and `actor_definitions` in that their Idempotency Key is compound with their `actor_type` -- not just another PK. This is needed to distinguish between the two types `source` and `destination`. Otherwise, you could conflict on a source create with a destination and vice versa. If there were separate tables for source/destination and their definitions, this would be less of an issue.

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_50_33_016__AddIdempotencyKeys.java
```diff
@@ -0,0 +1,59 @@
...
+    ctx.alterTable(table).addColumnIfNotExists(idempotencyKey).execute();
+    ctx.createUniqueIndex(indexName).on(table, columnName).execute();
+  }
+
+  private static void addActorIdempotencyKey(final @NotNull DSLContext ctx, final @NotNull String table) {
+    final String columnName = "idempotency_key";
+    final String indexName = String.format("%s_actor_type_idempotency_key", table);
+    final Field<UUID> idempotencyKey = DSL.field(columnName, SQLDataType.UUID.nullable(true));
+    ctx.alterTable(table).addColumnIfNotExists(idempotencyKey).execute();
+    ctx.createUniqueIndex(indexName).on(table, columnName, "actor_type").execute();
```
justenwalker, line 56: Perhaps `actor_type` should come first in the index instead of the `idempotency_key`; as the actor_type is sort of like a namespace for the `idempotency_key`.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SourceDefinitionsHandler.java
```diff
@@ -233,6 +238,20 @@ public SourceDefinitionRead getSourceDefinitionForWorkspace(final SourceDefiniti
...
+      Optional<StandardSourceDefinition> found = sourceService.getStandardSourceDefinitionByIdempotencyKey(idempotencyKey);
+      if (found.isPresent()) {
+        final StandardSourceDefinition sourceDefinition = found.get();
+        final ActorDefinitionVersion actorDefinitionVersion;
+        try {
+          actorDefinitionVersion = configRepository.getActorDefinitionVersion(sourceDefinition.getDefaultVersionId());
+          return buildSourceDefinitionRead(sourceDefinition, actorDefinitionVersion);
+        } catch (ConfigNotFoundException e) {
+          // not found, so must create
+        }
```
justenwalker, line 252: I'm not too happy with this. I suspect this is not a situation that you can get in, since the existence of the sourcedef implies it has at least one version (i would think). But i have to catch this due to `buildSourceDefinitionRead` throwing it.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/DestinationDefinitionsHandler.java
```diff
@@ -233,6 +238,20 @@ public DestinationDefinitionRead getDestinationDefinitionForScope(final ActorDef
...
+      final Optional<StandardDestinationDefinition> found = destinationService.getStandardDestinationDefinitionByIdempotencyKey(idempotencyKey);
+      if (found.isPresent()) {
+        final StandardDestinationDefinition destinationDefinition = found.get();
+        final ActorDefinitionVersion actorDefinitionVersion;
+        try {
+          actorDefinitionVersion = configRepository.getActorDefinitionVersion(destinationDefinition.getDefaultVersionId());
+          return buildDestinationDefinitionRead(destinationDefinition, actorDefinitionVersion);
+        } catch (ConfigNotFoundException e) {
+          // not found, so must create
+        }
```
justenwalker, line 252: I'm not too happy with this. I suspect this is not a situation that you can get in, since the existence of the destdef implies it has at least one version (i would think). But i have to catch this due to `buildDestinationDefinitionRead` throwing it.

PR #309, opened 2024-03-17
Use merge policy instead of fallbacks for KubeResourceConfig

azorej:
## What
Fix for https://github.com/airbytehq/airbyte/issues/33068

But the problem is more complex.
Logic in `io.airbyte.commons.workers.config.WorkerConfigsProvider`  is opaque.
The main strategy of resolving `KubeResourceConfig` is to fallback to the default for the whole config: if we cannot find the exact key, we will choose the closest default section.
However, there are some exceptions: annotations and `ResourceRequirements` (cpu/memory limits/requests). If any of these are empty, we will attempt to update them using the default config.

In addition, the current fallback logic does not work well for Helm OSS installations.
We parameterize config values via environment variables.
And there is no way to remove a section from the `application.yml` without using a custom image.
So one must setup all variables `JOB_KUBE_LABELS`, `CHECK_JOB_KUBE_LABELS`, `DISCOVER_JOB_KUBE_LABELS`, `REPLICATION_JOB_KUBE_LABELS`, `SPEC_JOB_KUBE_LABELS` in case they want to use common k8s labels (for example, `azure.workload.identity/use`).

The last issue I have is that I am not sure how to set a `null` value for a field in a `KubeResourceConfig` using environment variables.

## How
I suggest to introduce two breaking changes:
1. Use a merge strategy to resolve `KubeResourceConfig`. It's more intuitive and powerful.
2. Use the literal `<EMPTY>` to differentiate between empty and null fields.

The merge strategy will use the same order as previously: variant, type, subtype.
But it will try to merge configs based on `COALESCE(value, default_value)` for every field in `KubeResourceConfig` for every key part.
With this we don't need custom logic for annotations and `ResourceRequirements`.

`<EMPTY>` literal is needed to override defaults with empty values.

## Recommended reading order
1. `airbyte-commons-with-dependencies/src/test/resources/application-config-test.yaml`
2. `airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/WorkerConfigsProvider.java`
3. `airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/KubeResourceConfig.java`
4. all other files

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨

### DIFF
--- a/airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/WorkerConfigsProvider.java
+++ b/airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/WorkerConfigsProvider.java
@@ -282,7 +282,7 @@ public class WorkerConfigsProvider implements ResourceRequirementsProvider {
     if (defaultConfig.isEmpty()) {
       return variantConfig;
     }
-    return variantConfig.map(kubeResourceConfig -> defaultConfig.get().clone().update(kubeResourceConfig)).or(() -> defaultConfig);
+    return variantConfig.map(kubeResourceConfig -> kubeResourceConfig.merge(defaultConfig.get())).or(() -> defaultConfig);
   }
 
   private static Optional<KubeResourceConfig> getKubeResourceConfigByType(
@@ -301,7 +301,7 @@ public class WorkerConfigsProvider implements ResourceRequirementsProvider {
     if (defaultConfig.isEmpty()) {
       return typeConfig;
     }
-    return typeConfig.map(kubeResourceConfig -> defaultConfig.get().clone().update(kubeResourceConfig)).or(() -> defaultConfig);
+    return typeConfig.map(kubeResourceConfig -> kubeResourceConfig.merge(defaultConfig.get())).or(() -> defaultConfig);
   }
 
   private static Optional<KubeResourceConfig> getKubeResourceConfigBySubType(final Map<ResourceSubType, KubeResourceConfig> configBySubType,
@@ -319,7 +319,7 @@ public class WorkerConfigsProvider implements ResourceRequirementsProvider {
     if (defaultConfig.isEmpty()) {
       return subTypeConfig;
     }
-    return subTypeConfig.map(kubeResourceConfig -> defaultConfig.get().clone().update(kubeResourceConfig)).or(() -> defaultConfig);
+    return subTypeConfig.map(kubeResourceConfig -> kubeResourceConfig.merge(defaultConfig.get())).or(() -> defaultConfig);
   }
 
   private void validateIsolatedPoolConfigInitialization(final boolean useCustomNodeSelector, final Map<String, String> isolatedNodeSelectors) {
--- a/airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/WorkerConfigsProvider.java
+++ b/airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/WorkerConfigsProvider.java
@@ -165,7 +165,6 @@ public class WorkerConfigsProvider implements ResourceRequirementsProvider {
 
   @Singleton
   record WorkerConfigsDefaults(WorkerEnvironment workerEnvironment,
-                               @Named("default") KubeResourceConfig defaultKubeResourceConfig,
                                List<TolerationPOJO> jobKubeTolerations,
                                @Value("${airbyte.worker.isolated.kube.node-selectors}") String isolatedNodeSelectors,
                                @Value("${airbyte.worker.isolated.kube.use-custom-node-selector}") boolean useCustomNodeSelector,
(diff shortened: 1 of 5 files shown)


airbyte-commons-with-dependencies/src/test/resources/application-config-test.yaml
```diff
@@ -38,6 +39,7 @@ airbyte:
         cpu-request: 42
       micronauttest-source:
         cpu-limit: 5
+        cpu-request: <EMPTY> # Disable inheritance, force empty value
```
azorej, line 42: Be careful: that's a breaking change

airbyte-container-orchestrator/src/main/resources/application.yml
```diff
@@ -104,6 +104,7 @@ airbyte:
     kube-job-configs:
       default:
         annotations: ${JOB_KUBE_ANNOTATIONS:}
+        labels: ${JOB_KUBE_LABELS:}
```
azorej, line 107: Fix for https://github.com/airbytehq/airbyte/issues/33068

airbyte-commons-with-dependencies/src/test/resources/application-config-test.yaml
```diff
@@ -5,19 +5,20 @@ airbyte:
   worker:
     kube-job-configs:
       default:
+        annotations: default_annotation_key=default_annotation_value
         cpu-limit: default cpu limit
-        cpu-request: ${JOB_MAIN_CONTAINER_CPU_REQUEST:}
-        memory-request: ${JOB_MAIN_CONTAINER_MEMORY_REQUEST:}
+        cpu-request: default cpu request
+        memory-request: default memory request
```
azorej, line 11: Previous test cases weren't consistent

airbyte-commons-with-dependencies/src/test/resources/application-config-test.yaml
```diff
@@ -5,19 +5,20 @@ airbyte:
...
       check:
-        annotations: ${CHECK_JOB_KUBE_ANNOTATIONS:check annotations}
+        annotations: ${CHECK_JOB_KUBE_ANNOTATIONS:check_annotation_key=check_annotation_value}
```
azorej, line 13: We didn't check annotation overrides earlier

airbyte-commons-with-dependencies/src/test/java/io/airbyte/commons/workers/config/WorkerConfigProviderMicronautTest.java
```diff
@@ -71,11 +70,11 @@ void testKubeConfigIsReadingAllTheFields() {
...
   void testDefaultFieldBehavior() {
     assertEquals("spec", specKubeResourceConfig.getName());
-    assertEquals("spec annotations", specKubeResourceConfig.getAnnotations());
+    assertEquals("", specKubeResourceConfig.getAnnotations());
     assertEquals("spec labels", specKubeResourceConfig.getLabels());
     assertEquals("spec node selectors", specKubeResourceConfig.getNodeSelectors());
-    assertNull(specKubeResourceConfig.getCpuLimit());
-    assertNull(specKubeResourceConfig.getCpuRequest());
+    assertEquals("", specKubeResourceConfig.getCpuLimit());
+    assertEquals("", specKubeResourceConfig.getCpuRequest());
```
azorej, line 77: We consider nulls as empty strings for `KubeResourceConfig` now

airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/WorkerConfigsProvider.java
```diff
@@ -369,16 +369,12 @@ private Map<String, String> splitKVPairsFromEnvString(final String input) {
...
-        .withMemoryRequest(useDefaultIfEmpty(kubeResourceConfig.getMemoryRequest(), defaultConfig.getMemoryRequest()));
-  }
-
-  private static String useDefaultIfEmpty(final String value, final String defaultValue) {
-    return (value == null || value.isBlank()) ? defaultValue : value;
+        .withCpuLimit(kubeResourceConfig.getCpuLimit())
+        .withCpuRequest(kubeResourceConfig.getCpuRequest())
+        .withMemoryLimit(kubeResourceConfig.getMemoryLimit())
+        .withMemoryRequest(kubeResourceConfig.getMemoryRequest());
   }
```
azorej, line 378: Remove custom fallbacks since we don't need them anymore

airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/WorkerConfigsProvider.java
```diff
@@ -231,22 +230,13 @@ private WorkerConfigs getConfig(final KubeResourceKey key) {
...
 
-    // if annotations are not defined for this specific resource, then fallback to the default
-    // resource's annotations
-    final Map<String, String> annotations;
-    if (Strings.isNullOrEmpty(kubeResourceConfig.getAnnotations())) {
-      annotations = splitKVPairsFromEnvString(workerConfigsDefaults.defaultKubeResourceConfig.getAnnotations());
-    } else {
-      annotations = splitKVPairsFromEnvString(kubeResourceConfig.getAnnotations());
-    }
-
```
azorej, line 242: Remove custom fallbacks since we don't need them anymore

mhemken-vts:
Can ALL the resource settings be exposed please?
* [destination-stdin][dststdin]
* [source-stdout][srcstdout]

[dststdin]: https://github.com/airbytehq/airbyte-platform/blob/f2819ce4a990c783581cdc770c39ebed6161d800/airbyte-server/src/main/resources/application.yml#L189 "stdin"
[srcstdout]: https://github.com/airbytehq/airbyte-platform/blob/f2819ce4a990c783581cdc770c39ebed6161d800/airbyte-server/src/main/resources/application.yml#L204 "stdout"

PR #311, opened 2024-03-19
Updated for tcp6 support.

xorl:
## What
socat only binds to TCP4 on workers.

## How
socat only binds to TCP4 on worker nodes so we modify it to work with TCP6, the socat TCP6 binds to both TCP4 and TCP6 so support native IPv6 only clusters.

## Recommended reading order
1. `airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java`

*If unsure, leave it blank.*
- [x] YES 💚
- [ ] NO ❌

## 🚨 User Impact 🚨
May cause issues if IPv6 routing is not configured correctly.


xorl:
Any update on this?

PR #318, opened 2024-03-29
Fix document link

kiwamizamurai:
## What

Make links of kubernetes document valid

## How

Replace old url with new url

## Recommended reading order
1. `x.kt`
2. `y.kt`

- [x] YES 💚
- [ ] NO ❌


marcosmarxm:
Thanks for the contribution @kiwamizamurai there was a large change in values file and it will hard to solve conflicts. I'll share this to the deployment team make sure the URL are updated too.

PR #319, opened 2024-04-08
Configurable TIMEOUTs for Worker via ENV VAR

hanslemm:
## What
This PR introduces configurable Timeouts for Worker constants via ENV VARs.
For some K8s setup where nodes are spun up on demand, it might be the case that 2 minutes tolerance for pod creation is not enough.

## How

With the what in mind, this PR introduces the following ENV VARs that maps to their correspondent constants:
- `INIT_CONTAINER_STARTUP_TIMEOUT` -> accepts only >0 values.
- `INIT_CONTAINER_TERMINATION_TIMEOUT` ->  accepts only >0 values.
- `POD_READY_TIMEOUT` -> accepts only >0 values.

When the ENV VARs are not set, the current default values are assumed.

Furthermore, when the user provides an invalid value for such constants, the `parseEnvVar` function handles it, printing a helpful message saying that the format is invalid, and that it is assuming the default value.

- [X] YES 💚
- [ ] NO ❌


bgroff:
The default timeout has been adjusted to 15 minutes per attempt. We will try 5 attempts, so this is roughly an hour and a half. With this information, I am going to close this PR as wont fix.

david-freistrom:
@bgroff 

> The default timeout has been adjusted to 15 minutes per attempt. We will try 5 attempts, so this is roughly an hour and a half. With this information, I am going to close this PR as wont fix.

Where you defined 15 minutes? The code which is refereced in the PR  shows 2 minutes. And there is no attempt for such pod creation at all.

`public static final Duration POD_READY_TIMEOUT = Duration.ofMinutes(2);`

2 Minutes is for serverless arch and autoscaled nodes not enough if we first need to provide a compute resources.

PR #320, opened 2024-04-08
Add an optional endpoint to S3 configs

francois-travais:
## What

I would like to be able to use Airbyte OSS with a S3-compatible storage for logs. The only missing piece is to be able to give a custom endpoint to S3 clients.

## How

I add an optionnal argument, `endpoint` to the S3 storage

## Recommended reading order
1. `CloudStorageConfigs.java`

- [x] YES 💚
- [ ] NO ❌


francois-travais:
closes https://github.com/airbytehq/airbyte/issues/36896

mhkarimi1383:
@francois-travais 

Can you also add an option to set ForcePathStyle?

PR #322, opened 2024-04-16
Fix opening non default workspace in webapp

nathanseal:
The problem was that opening the webapp to a workspace id that was not the default worksapce id was failing with the `Invalid RBAC query` error.

It seems like it was introduced in this change:
https://github.com/airbytehq/airbyte-platform/commit/61079f598c249eb098147a1f5fba096650683a05#diff-aa38ef9df589b86ae09c4ac9c1653be68a79ed26c07dcb0a75d641de00c99b03R22-R26

Simply reverting the extra check that was added seems to fix the issue.

## Recommended reading order
1. `x.kt`
2. `y.kt`

- [x] YES 💚
- [ ] NO ❌


msaffitz:
We're seeing this as well in the latest builds & helm charts.  Would be great to get this merged!  Thank you!

cayden-uw:
@timroes since he looks like the original commiter of the change @nathanseal mentioned

timroes:
Hi, the issue is solved from version [v0.57.4](https://github.com/airbytehq/airbyte-platform/releases/tag/v0.57.4) onwards. Please upgrade to that version to fix the issue. 

_Background:_ Every workspace should be in an organization (data model wise) by now, and that it wasn't (in the OSS version) was due to some issues, which have now been addressed. The code in question here in this PR (specifically: the exception that there wouldn't be an organization) will be removed soon completely.

cayden-uw:
Thanks @timroes!

PR #329, opened 2024-05-14
Add nodeSelector, tolerations, and affinity to minio create bucket pod

jprpai:
## What
The minio create bucket pod does not inherit the tolerations and affinities the main minio does. We need the ability to define where this pod runs. In our install this pod does not run and the helm chart leaves an error condition. 

## How
The create bucket pod uses the same settings as the other minio pods.

## Recommended reading order
1. `x.kt`
2. `y.kt`

- [ X ] YES 💚
- [ ] NO ❌

### DIFF
--- a/charts/airbyte/templates/minio.yaml
+++ b/charts/airbyte/templates/minio.yaml
@@ -164,4 +164,16 @@ spec:
             configMapKeyRef:
               name: {{ .Release.Name }}-airbyte-env
               key: MINIO_ENDPOINT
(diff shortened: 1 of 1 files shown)


abusutil:
@octavia-approvington Anyone from Airbyte team could be check this PR please ? Could be important to have this for multi-architecteur kubernetes cluster.

Thanks in advance

IIAzAII:
Can this get a bump because this would be extremely useful!

PR #331, opened 2024-05-14
Update OrchestratorConstants.kt

ro8inmorgan:
Expose SIDECAR env variables to orchestrator

## What
Expose SIDECAR env variables to orchestrator

## How
Added the correct variables

- [ X] YES 💚
- [ ] NO ❌


ro8inmorgan:
Why is it closed?

PR #333, opened 2024-06-13
fix(socat): handle ipv6 and ipv4 communication from socat to runner

ogirardot:
## What
When trying to use Airbyte in an IPV6 enabled kubernetes cluster the socat process fails with 
```
socat[8] N reading from and writing to stdio
socat[8] E TCP: wrong number of parameters (9 instead of 2)
socat[8] N exit(1)
```

This PR fixes that using the TCP6:[IP]:port syntax when identifying an IPV6 address for the original process runner host.

## How
By checking the process runner host to check if it is an Ipv6 or Ipv4 format

- [X] YES 💚
- [ ] NO ❌


xorl:
+1 on this, I wrote a patch for internal testing twice now on two separate pull requests. This needs more attention and a fix so we do not have to manually patch it internally.

PR #335, opened 2024-06-25
image secrets property added to asyc pod

sivankumar86:
## What
    We are running airbyte in private EKS hence, we need to add private repo to download the images from docker which is required to add imagesecret whilst launching pod.

## How
  Added image secret from job main imagesecret 

https://github.com/airbytehq/airbyte/issues/38177

https://github.com/airbytehq/airbyte/issues/33620


sivankumar86:
https://github.com/airbytehq/airbyte/issues/42091

butterfly-thomas:
Same error here with Azure AKS. I tried the newly release v1.0.0 but get and 500 error instead... Any update?

PR #351, opened 2024-08-21
Fix incorrect usage of Go templating

XIAZY:
## What

The usage of Go templating (`index` function more specifically) was incorrect in recent releases.

Instead of populating configmap items `CONTAINER_ORCHESTRATOR_IMAGE`, `WORKLOAD_INIT_IMAGE` and `CONNECTOR_SIDECAR_IMAGE` with the right value from their respective `image` field, the helm chart populates them as `null`, which is incorrect. From Go documentation, 
> index
	Returns the result of indexing its first argument by the
	following arguments. Thus "index x 1 2 3" is, in Go syntax,
	x[1][2][3]. Each indexed item must be a map, slice, or array.

This fix has solved the problem for me locally
## How

The fix replaces incorrect usage of `index` with nested indices separated by spaces
## Recommended reading order
1. `x.kt`
2. `y.kt`

- [ ] YES 💚
- [ ] NO ❌


marcosmarxm:
@XIAZY sorry the delay to review your contribution.
@perangel and @abuchanan-airbyte can you check this?

perangel (approved):

marcosmarxm:
This was merged in https://github.com/airbytehq/airbyte-platform-internal/pull/14077

PR #355, opened 2024-09-03
Fix helm template for `airbyte-workload-api-server` (GCS logging)

brianstorti:
Even when we have a secret with the gcs credentials (as described in this [doc](https://docs.airbyte.com/deploying-airbyte/creating-secrets)), this template is still trying to use `{release}-gcs-log-creds` instead of what we have configured in `.Values.global.storage.storageSecretName`.

I first noticed this issue when trying to upgrade to chart `0.429.0`, as it's not creating the `{release}-gcs-log-creds` anymore and the `airbyte-workload-api-server` deployment fails with this error:
```
MountVolume.SetUp failed for volume "gcs-log-creds-volume" : secret "airbyte-gcs-log-creds" not found
```


marcosmarxm:
@airbytehq/platform-deployments can someone take a look?

perangel:
@brianstorti There is project underway to clean up a number of things in the chart, this issue is one of them. The general solution (which is to set the secret name based on whether one is specified by the user or not) is the right answer. If you want to get this is as a standalone fix, that would be welcome, but you'll need to update all places where that secret is mounted (unless this is the only workload that currently is setting it incorrectly). 

Those should be:
- `charts/airbyte-featureflag-server/templates/deployment.yaml`
- `charts/airbyte-seerver/templates/deployment.yaml`
- `charts/airbyte-worker/templates/deployment.yaml`
- `charts/airbyte-workload-api-server/templates/deployment.yaml`
- `charts/airbyte-workload-launcher/templates/deployment.yaml`

I believe part of the issue is that the way its specified across those various templates is inconsistent.

perangel:
Looks like this is the outstanding template that needs correcting. Going to test this out internally and see if we can't get this merged in today

perangel:
This has been merged! Thanks for the contribution!

PR #369, opened 2024-11-06
fix: add tolerations during reset pod creation

SamyDjemai:
## What
While replication pods took tolerations into consideration, reset pods were created without tolerations, which resulted in Pending pods due to node taints.

## How
Create reset pods with provided tolerations, just like replication pods. Closes https://github.com/airbytehq/airbyte/issues/48380

## Recommended reading order
1. `airbyte-workload-launcher/src/main/kotlin/pods/factories/ReplicationPodFactory.kt`

- [x] YES 💚
- [ ] NO ❌


abuchanan-airbyte:
Applied in https://github.com/airbytehq/airbyte-platform/commit/84a7536639fceda98cb923e832369ed533f734cf

Sorry, I wasn't familiar with the process for how we apply changes from PR to the platform (it's not a simple merge), so the commit shows up under my name.

SamyDjemai:
No problem, thanks!

PR #370, opened 2024-11-06
fix: ENV VAR kv parsing to handle json values

bHacklv:
## What
- JOB_KUBE_ANNOTATIONS can have JSON as value
- This fixes the parsin of such values

### Before fix:
Image

### After fix:
See new UTs

After fix:

## How

Example of a valid istio k8s annotation that is incorrectly parsed without a fix:
```
proxy.istio.io/config={ "holdApplicationUntilProxyStarts": true, "proxyMetadata": { "EXIT_ON_ZERO_ACTIVE_CONNECTIONS": "true" } }
```

- [ ] YES 💚
- [ ] NO ❌


airbyte-commons-with-dependencies/src/main/java/io/airbyte/commons/workers/config/EnvUtils.java
colesnodgrass, line 1: Can we do this in kotlin instead of java?  I don't think we should be introducing new java classes at this point.

colesnodgrass:
This PR will need to be updated against the `main` branch as it has dependencies which no longer exist within the platform (i.e. lombok).

bHacklv:
> This PR will need to be updated against the `main` branch as it has dependencies which no longer exist within the platform (i.e. lombok).

Sure, I will do it on Monday, thanks!

bHacklv:
> This PR will need to be updated against the `main` branch as it has dependencies which no longer exist within the platform (i.e. lombok).

Finally got time to remove the usage of lombok.
It's ready for re-review!

PR #386, opened 2025-01-31
SELECT specific columns from ORGANIZATION table in getOrganization

HaJunYoo:
## What

During Airbyte helm upgrade from 1.2.0 to 1.4.0, a type mismatch error occurred between code and database schema, causing web server access failure.

Root Cause
- Organization table's `tombstone` column contains UUID values
- Code attempts to read this field as boolean type
- Type conversion failure triggers 500 error on `/api/v1/instance_configuration` endpoint
- This occurs during organization data fetch operations

Migration Issue
- Database schema changes were required but not properly applied
- Flyway migration V0_50_24_003 should have added new columns:
   - 'pba' column
   - 'org_level_billing' column
- Incomplete migration led to schema inconsistency in tombstone column

Web server became inaccessible due to the failed type conversion between UUID and boolean in the organization table.

Error message:
```
Caused by: org.postgresql.util.PSQLException: Cannot cast to boolean: "4f74841b-3fa8-476c-9102-67a80be8a042"
at org.postgresql.jdbc.BooleanTypeUtil.cannotCoerceException(BooleanTypeUtil.java:99)
```

## How
Modified organization queries to explicitly select required columns instead of using asterisk(*):

Removed select(asterisk()) to avoid automatic mapping of problematic tombstone field
Added explicit column selection for required fields only
This prevents the type mismatch between UUID and boolean

## Recommended reading order

- OrganizationPersistence.java

- [x] YES 💚
- [ ] NO ❌

### DIFF
--- a/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/OrganizationPersistence.java
+++ b/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/OrganizationPersistence.java
@@ -58,10 +58,18 @@ public class OrganizationPersistence {
    */
   public Optional<Organization> getOrganization(final UUID organizationId) throws IOException {
     final Result<Record> result = database.query(ctx -> ctx
-        .select(asterisk())
-        .from(ORGANIZATION)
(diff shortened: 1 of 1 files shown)


cgardens:

PR #387, opened 2025-02-04
🐛 Fix undefined prefix in stream name

szemek:
## What

There is a bug in Airbyte 1.4.0 that for some connections it display stream name with "undefined" prefix as in attached screenshot.

image

## How

It uses previous implementation as here: https://github.com/airbytehq/airbyte-platform/commit/ea7af1f9bd1adfee1471cd6c32349ed4c501ba5c
It sets prefix conditionally. For falsy values, it sets empty string "" as a prefix.

## Recommended reading order

1. `airbyte-webapp/src/area/connection/utils/useUiStreamsStates.ts`

- [x] YES 💚
- [ ] NO ❌


josephkmh:
@szemek thank you for the fix 🎉 

This is merged in our upstream repository and will go out in the next Airbyte release.

PR #388, opened 2025-02-20
Removed unused externalDatabase Helm chart configuration

Tybot204:
## What

Issue: https://github.com/airbytehq/airbyte/issues/54176

Current Helm `values.yml` documentation and defaults have old unused values for `externalDatabase`, potentially leading to confusion when setting up or upgrading Helm chart versions.

## How

This removes `externalDatabase` from all defaults and documentation. These values are not used anywhere in the project outside of the changes files in this PR. Instead `global.database` should be used instead to configure external PostgreSQL connections.

- [x] YES 💚
- [ ] NO ❌


marcosmarxm:
cc @airbytehq/platform-deployments

PR #396, opened 2025-04-15
🐛 Fix Helm Charts: add a storage class name for Min IO

saifsafsf:
## What
Adding a storage class label in helm charts for airbyte with Min IO because without no storage class name & no default set, it throws this error.
`no persistent volumes available for this claim and no storage class is set`

## How
A section has been added in the minio.yaml file in the airbyte templates where the storageClassName would be set only if a value has been passed.

## Recommended reading order
1. `minio.yaml`
2. `values.yaml`

- [x] YES :green_heart:
- [ ] NO :x:


saifsafsf:
Replaced with #407

PR #397, opened 2025-04-15
🐛 Fix Helm Charts: place volumes: inside the if condition

saifsafsf:
## What
`volumes:` label  in `spec.template.spec.containers` field throws the following warning if no `extraVolumes` are passed.
`warnings.go:70] unknown field "spec.template.spec.containers[0].volumes`

## How
Placing the `volumes:` label inside the if condition helps avoid the aforementioned warning when no `extraVolumes` are passed.

- [x] YES :green_heart:
- [ ] NO :x:


saifsafsf:
Replaced with #404

PR #398, opened 2025-04-15
🐛 Fix Helm Charts: remove leaking database credentials from airbyte's config map

saifsafsf:
## What
The `airbyte_env` config map in `airbyte` charts includes database credentials, which should be a secret instead.

## How
Remove the credentials from the `airbyte_env` config map since these secrets are not referred to anywhere through this config map.

- [x] YES :green_heart:
- [ ] NO :x:


saifsafsf:
Replaced with #405

PR #401, opened 2025-04-22
for issue airbytehq/airbyte#55172, add pvc storage capacity to helm chart

ufou:
## What
Allow a helm chart parameter to override the capacity of the postgres db pvc

## How
Add a default parameter for the postgres db pvc capacity in `values.yaml` and use that value as a variable in the template, instead of the current hard coded value.

- :white_check_mark: YES 💚


cgardens:

PR #418, opened 2025-07-14
chore: Remove errant Workspace#list entity operation

bflad:
## What

The expected entity operations for downstream tooling are one of `create`, `read`, `update`, and `delete`:

```
Error: Unexpected op annotation list applied as Workspace#list: expected it to be one of ["create","read","update","delete"]
```

Since the `listWorkspaces` API operation appears to be intended as its own Terraform data resource and its usage would be separate from the workspace managed resource, the entity operation could be updated to create a workspaces data resource via `Workspaces#read`. However, it is removed for now to match the current behavior of not doing anything in the Terraform code generation while also preventing the error.

## How

Removed various `x-sdk-entity-operation: Workspace#list` and `x-speakeasy-entity-operation: Workspace#list` usage in OpenAPI Specification documents.

## Recommended reading order
1. `x.kt`
2. `y.kt`

- [ ] YES 💚
- [ ] NO ❌

### DIFF
--- a/airbyte-api/server-api/src/main/openapi/api.yaml
+++ b/airbyte-api/server-api/src/main/openapi/api.yaml
@@ -2089,7 +2089,6 @@ paths:
       x-speakeasy-alias: listWorkspaces
       x-speakeasy-group: Workspaces
       summary: List workspaces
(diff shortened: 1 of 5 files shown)


cgardens:

PR #438, opened 2026-03-18
fix(config-secrets): fix Google Secret Manager regional secrets and add CMEK support

erik-astranis:
## What

Fixes airbytehq/airbyte#75191

Google Secret Manager regional secrets are partially implemented but broken in several ways, preventing self-hosted deployments from using regional GSM.

Additionally I bumped the `google-cloud-secretmanager` dependency to `2.76.0` and it seems that it was already being tranitively bumped via `micronaut` anyways.

```bash
./gradlew :oss:airbyte-config:config-secrets:dependencyInsight \
  --dependency google-cloud-secretmanager --configuration runtimeClasspath
```

## How

Three fixes in `GoogleSecretManagerPersistence.kt`:

1. **Missing version suffix**: `getSecret()` built regional resource names without `/versions/latest`, so `accessSecretVersion()` was handed a secret rather than a secret version. Added the suffix.

2. **Incompatible replication policy**: `getReplicationPolicy()` always returned `Automatic`, but regional secrets reject a replication policy. It now returns `null` for regional secrets so no policy is set.

3. **CMEK support**: Both Regional and Global are implemented.

Both constraints are per Google's docs ([CMEK](https://cloud.google.com/secret-manager/docs/cmek), [regional CMEK](https://cloud.google.com/secret-manager/regional-secrets/create-secret-cmek-encryption)).

## Questions

Helm chart exposes `kmsKeyName`, but it doesn't seem like I can re-render it from this repo. Happy to do it with some guidance.

Separately, Claude claims there is a pre-existing bug in `GoogleSecretManagerPersistence.kt`: `disableAllSecretVersions()` passes `secretParent()` (a project, or project/location) as the `ListSecretVersionsRequest` parent, where that endpoint expects the secret resource name. Left it alone as unrelated.

## Review guide

1. `GoogleSecretManagerPersistence.kt` — single file, all changes

## User Impact

Self-hosted deployments using regional Google Secret Manager will be able to read and create secrets correctly. CMEK encryption is now configurable for compliance requirements.

- [x] YES 💚


erik-astranis:
Working on a refresh

erik-astranis:
Refresh done.

